PR #57, opened 2015-03-20
Check if class woo commerce exists [Delivers #87198510]

### LINKED ISSUE #53: Check if WooCommerce is active
When the WooCommerce plugin is not active and Yoast WooCommerce SEO plugin is activated, there is a fatal error on the front end.

schermafbeelding 2015-01-29 om 14 41 06

There should be a check if WooCommerce is installed and activated, if not, do not load the plugin at all. Ideally, there should be a notice to inform the user the plugin couldn't be activated as WooCommerce isn't active.

andizer:
Fixes #53


omarreiss:
CR :+1:

omarreiss:
Acceptance :white_check_mark:

Issue #77, opened 2015-11-05
'The short description for this product is too short' on a non Woo page


CarolineGeven:
The notification 'the short description for this product is too short.' appears on a post page. This page analysis notification is Woo specific and should only show up on the product pages of Woo.

WP 4.3 - Yoast SEO 3.0 - 7, Yoast SEO WooCommerce trunk

PR #81, opened 2015-11-13
Fix fatal JS error when excerpt is not present

atimmer:
Fixes https://github.com/Yoast/wordpress-seo/issues/3187
Also fixes https://github.com/Yoast/wpseo-woocommerce/issues/77


terw-dan:
CR Done 👍

diedexx:
Acceptance done :+1:

PR #88, opened 2016-01-15
Allow for the new wp_installing() function.

jrfnl:
Introduced in WP 4.4.
See: https://core.trac.wordpress.org/changeset/34828

### DIFF
--- a/wpseo-woocommerce.php
+++ b/wpseo-woocommerce.php
@@ -822,7 +822,7 @@ function initialize_yoast_woocommerce_seo() {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


atimmer:
The `if` in this change has now become a monstrosity, I would like to suggest we mock the `wp_installing` function that checks for `WP_INSTALLING` using `function_exists` and then this if just becomes:
```php
if ( ! wp_installing() ) {
```

jrfnl:
See my response https://github.com/Yoast/wpseo-video/pull/213#issuecomment-178661034

atimmer:
See https://github.com/Yoast/wpseo-woocommerce/pull/92

PR #90, opened 2016-01-21
Do not cut hand-crafted post excerpts

NickIvanter:
when forming the meta- and og:description. Only use the hard 156-char limit when there is no excerpt defined by the user and we have to cut a piece from the entire post content.

### DIFF
--- a/wpseo-woocommerce.php
+++ b/wpseo-woocommerce.php
@@ -590,10 +590,10 @@ class Yoast_WooCommerce_SEO {
 						$metadesc = $product->post->post_excerpt;
 					} elseif ( $product->post->post_content != '' ) {
 						$metadesc = $product->post->post_content;
-					}
 
-					if ( ! empty( $metadesc ) ) {
-						$metadesc = wp_html_excerpt( $metadesc, 156 );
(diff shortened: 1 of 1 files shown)


wpseo-woocommerce.php
```diff
@@ -590,10 +590,10 @@ public function metadesc( $metadesc ) {
 						$metadesc = $product->post->post_excerpt;
 					} elseif ( $product->post->post_content != '' ) {
 						$metadesc = $product->post->post_content;
-					}
 
-					if ( ! empty( $metadesc ) ) {
-						$metadesc = wp_html_excerpt( $metadesc, 156 );
+						if ( ! empty( $metadesc ) ) {
```
moorscode: This is redundant and the logic can be combined into one line by using the `$product->post->post_content` as argument of `wp_html_excerpt`.

moorscode:
Hi @NickIvanter,

Thank you for noticing this and taking the time to submit an issue including a fix.
Could you do a minor change to the code so we can merge it after acceptance testing it?

jcomack:
Closing this in favor of #171

PR #92, opened 2016-02-09
Use wp_installing

atimmer:
wp_installing() is now the correct way to check if WordPress is
installing.

See https://core.trac.wordpress.org/changeset/34828


CarolineGeven:
CR done

PR #96, opened 2016-02-25
Activate license on plugin activation.

andizer:
When the plugin is activated we should try to activate the license as well.


omarreiss:
CR :+1:

terw-dan:
Acceptance :+1:

PR #99, opened 2016-02-29
Fix price og tags

jdevalk:
Our OpenGraph tags for products are wrong, this fixes it.

This needs _extensive_ testing with Facebook and Pinterest, as [Pinterest's documentation](https://developers.pinterest.com/docs/rich-pins/products/) for this is... Different from [Facebook's](https://developers.facebook.com/docs/reference/opengraph/object-type/product/). 

If this code doesn't work with Pinterest, we should reach out to them and/or make a workaround for them too.


omarreiss:
For acceptance: Make sure this still works for pinterest as well. Use the pinterest validator.

terw-dan:
Acceptance: 

I get a notice on the WooCommerce SEO Settings page: 
Notice: Undefined variable: i in /srv/www/wordpress-default/wp-content/plugins/wpseo-woocommerce/wpseo-woocommerce.php on line 306

terw-dan:
Acceptance.

The tags in pinterest seem to be working fine, in the pinterest debugger it shows all tags once. 

The facebook debugger tool still seems to find a few problems with the new og tags. 

old:
image

new: 
image

@jdevalk is this the intended behaviour? if so it can be merged.

jdevalk:
ah. No. The page type needs to be changed. Put it back to development :)

jdevalk:
Ok, fixed. Tested myself in Pinterest and FB and should be good now, but please review code again :)

jcomack:
CR done :+1:

terw-dan:
Acceptance:
@jdevalk  with the new fix I don't have the warnings in facebook anymore, but when I check with the pinterest checker it shows all properties twice

image

with the old fix it only showed them once.
image

if this is intended behaviour the acceptance is OK and this can be merged, but I'd like to double check this.

PR #112, opened 2016-04-19
Fixed "update plugin" on the plugins page

moorscode:
Changed arguments to comply with Licence Manager updates so we use an HTML ID safe string for slug needed for correct usage of the shiny updates routine.

See https://github.com/Yoast/wordpress-seo-premium/issues/274

Consideration:
We realise that using an older License Manager version might break functionality but upgrading Free or Premium will solve this (which everybody should do), so not considered a huge problem.


jcomack:
CR done 👍

atimmer:
Acceptance done 👍🏻

PR #171, opened 2017-09-27
Don't cut hand-crafted excerpts

jcomack:
This code cleans up some of the code in the `metadesc` function and implements a fix introduced in #90.

Changelog entry:
* Fixes a bug where the short description was cut to the hard 156-character limit.

Props to @NickIvanter for the original fix.

### DIFF
--- a/wpseo-woocommerce.php
+++ b/wpseo-woocommerce.php
@@ -678,25 +678,25 @@ class Yoast_WooCommerce_SEO {
 			return '';
 		}
 
-		$product_id = get_the_id();
-		$product = $this->get_product_for_id( $product_id );
+		$product = $this->get_product_for_id( get_the_id() );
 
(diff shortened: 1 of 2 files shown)


wpseo-woocommerce.php
```diff
@@ -678,25 +678,25 @@ public function metadesc( $meta_description ) {
...
 
-		return '';
+		$short_description = $this->get_short_product_description( $product );
+		$long_description = $this->get_product_description( $product );
+
+		if ( $short_description !== '' ) {
+			$meta_description = $short_description;
+		}
+
+		if ( $long_description !== '' ) {
```
moorscode: When short_description is not empty, it should be respected and returned as `to-be-used` copy.
Long description will override short description in the proposed situation.

PR #198, opened 2018-02-07
Build: Travis tweaks

jrfnl:
## Summary
_N/A_

This PR contains two tweaks to the Travis build script - each in a separate commit for easier review. Also see the commit messages for additional info:

1. On line 50 of the script, a specific PHPUnit version is pulled in via Composer for certain PHP versions. This PHPUnit version is subsequently never used.
    This commit brings the command to run PHPUnit in line with this.
    Alternatively, - as it looks like the build runs fine with PHPUnit 6 -, the `composer require` on line 50 could be removed.
2. Validate the `composer.json` file during the builds.

## Test instructions
* Check the Travis build logs to see the effects.


jcomack:
CR & acceptance done 👍

PR #206, opened 2018-02-08
removed deprecated functions from wpseo_woocommerce

Dieterrr:
## Summary

*Removed all deprecated functions that were deprecated since before version 4.0 and were showing a warning message with '_deprecated_function' from woocommerce. One function was deleted despite not calling _deprecated_function, because it was completely empty.

## Test instructions

* This PR cannot be tested specifically. It's a code cleanup and it should have no effect on the plugin. The only difference is that developers will not see warning messages anymore from functions that have been deprecated since before 4.0. The only way of testing is checking that the plugin still works. 

## Quality assurance

* [x ] I have tested this code to the best of my abilities
--> I have run the excisiting unittests and they don't throw any errors. 

Fixes #8783


jrfnl:
Duplicate of merged PR #199. Suggest close.

> One function was deleted despite not calling _deprecated_function, because it was completely empty.

That function was only recently deprecated, so should not be removed (yet).


> The only difference is that developers will not see warning messages anymore from functions that have been deprecated since before 4.0. 

The only reason warning messages would be thrown, would be that those deprecated functions are actually being called. If that's the case, a backtrace is needed to figure out what plugin is calling those methods and things should be fixed there.

Dieterrr:
You're right. I'll add a deprecated message in in the coming days and will set it to deprecated 7.0. 
I'll close the PR.

jrfnl:
> I'll add a deprecated message in in the coming days and will set it to deprecated 7.0.

@Dieterrr No need. Already done in https://github.com/Yoast/wpseo-woocommerce/pull/210/commits/f3583b20202a75cb87528f6691c571d294aea5c2 which is part of PR #210.

PR #211, opened 2018-02-09
Add WordPress SEO version requirement

### LINKED ISSUE #195: Add hard version requirement of Yoast SEO 7.0
Because of no guarantees around upgrading order and making sure the methods are available in runtime, we need to force all customers to make sure they have 7.0 installed on Yoast SEO or Yoast SEO Premium.

This means adding a hard version requirement check for 7.0 and disabling the functionality (not the plugin) when the version is lower (using the `WPSEO_VERSION` constant).

### LINKED ISSUE #212: Update minimal required version for WordPress
In #211 the minimal version of WordPress SEO is updated. We should do the same for the WordPress version. This should be 4.8 instead what it is now.

andizer:
## Summary

* Show a notice when a WordPress SEO version below 7.0 is installed.
* Show a notice when a WordPress version below 4.8 is installed.

## Test instructions

* Have this branch and WordPress SEO active. Change in WordPress SEO the value of constant `WPSEO_VERSION` in a value below 7.0
* A notice should be shown. 
* Change the value of the constant to 7.0 or 7.1 and see the notice being gone.

* The minimal required WordPress version can be tested by adding this to the 'wpseo-woocommerce.php' file:
```php
add_action( 'admin_init', function() {
	global $wp_version;

	$wp_version = 4.7;
} );
```

Fixes #195 
Fixes #212

### DIFF
--- a/wpseo-woocommerce.php
+++ b/wpseo-woocommerce.php
@@ -67,7 +67,7 @@ class Yoast_WooCommerce_SEO {
 	}
 
 	/**
-	 * Class constructor
+	 * Class constructor.
 	 *
 	 * @since 1.0
 	 */
--- a/wpseo-woocommerce.php
+++ b/wpseo-woocommerce.php
@@ -102,8 +102,8 @@ class Yoast_WooCommerce_SEO {
 			return false;
 		}
 
-		// When version is below 7.0.
-		if ( ! version_compare( $wordpress_seo_version, '7.0', '>=' ) ) {
+		// Make sure Yoast SEO is at least 7.0, including the RC versions, so bigger than 6.9.
+		if ( ! version_compare( $wordpress_seo_version, '6.9', '>' ) ) {
 			add_action( 'all_admin_notices', 'yoast_wpseo_woocommerce_upgrade_error' );
 
 			return false;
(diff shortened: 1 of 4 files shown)


wpseo-woocommerce.php
```diff
@@ -67,11 +67,72 @@ public static function get_plugin_file() {
...
+
+	/**
+	 * Checks the dependencies. Sets a notice when requirements aren't met.
+	 *
+	 * @param string $wp_version The current version of WordPress.
+	 *
+	 * @return bool True whether the dependencies are okay.
+	 */
+	protected function check_dependencies( $wp_version ) {
+		if ( ! version_compare( $wp_version, '3.5', '>=' ) ) {
```
jrfnl: Should this be updated to 4.8 ? Which AFAICS is the official minimum supported WP version ?

andizer: Will open a new issue for that.

andizer:
I've change the minimal required WordPress version in this pull, because of I need the logic.

wpseo-woocommerce.php
```diff
@@ -67,11 +67,72 @@ public static function get_plugin_file() {
...
+
+		// When WordPress SEO is not installed.
+		if ( ! $wordpress_seo_version ) {
+			add_action( 'all_admin_notices', 'yoast_wpseo_woocommerce_missing_error' );
+
+			return false;
+		}
+
+		// When version is below 7.0.
+		if ( ! version_compare( $wordpress_seo_version, '7.0', '>=' ) ) {
```
moorscode: > 6.9

PR #249, opened 2018-05-15
Exclude artifact folder from pot file creation

moorscode:
## Summary

If an artifact was build previously, this will be taken into account when re-generating the language files. Resulting in double entries which are not relevant.

* _N/A_

* Excluded the `optionally present` artifact folder from being parsed for translations.

## Test instructions

* `grunt copy:artifact` - This generates an `artifact` folder in your local directory
* `grunt build:i18n` - This rebuilds the `pot` file

Related https://github.com/Yoast/wpseo-news/issues/406


jcomack:
CR & Acceptance done 👍

PR #259, opened 2018-06-19
Filters excluded catalog products from the sitemap

### LINKED ISSUE #42: Do not include hidden products in the sitemap
We received the following feedback from a customer:
"In Woocommerce you have the option to hide products from the Catalog.
From my understanding these hidden products also should NOT be included in the in the XML Sitemap that is generated by the Yoast SEO (Woocommerce) Plugin."

The option to hide a product from the catalog can be found in the publish-box on the edit-product-page.It's called Catalog visibility.

The customer adds:
"There is also a ready to use meta-query to query only visible products:
WC()->query->visibility_meta_query()
It is defined on line 595 in class-wc-query.php"

andizer:
## Summary

* Hides excluded catalog products from the products sitemap.

## Test instructions

* Add a product (for example: Product A)
* View the products sitemap and verify the product being visible.
* Edit the product by excluding it from the catalog:
cursor_and_edit_product_ _local_wordpress_dev_ _wordpress
edit_product_ _local_wordpress_dev_ _wordpress
* Visit the sitemap and the product should be hidden.

Fixes #42

### DIFF
--- a/tests/doubles/woocommerce-seo-double.php
+++ b/tests/doubles/woocommerce-seo-double.php
@@ -20,4 +20,11 @@ class Yoast_WooCommerce_SEO_Double extends Yoast_WooCommerce_SEO {
 	public function check_dependencies( $wp_version ) {
 		return parent::check_dependencies( $wp_version );
 	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function filter_hidden_product( $url, $type, $post ) {
+		return parent::filter_hidden_product( $url, $type, $post );
+	}
 }
--- a/tests/doubles/woocommerce-seo-double.php
+++ b/tests/doubles/woocommerce-seo-double.php
@@ -22,7 +22,7 @@ class Yoast_WooCommerce_SEO_Double extends Yoast_WooCommerce_SEO {
 	}
 
 	/**
-	 * @inheritdoc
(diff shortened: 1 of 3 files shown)


andizer:
@Dieterrr The unit tests are here: https://github.com/Yoast/wpseo-woocommerce/pull/259/commits/b41896f39e575cc48fb5f839ba6f6187f86051cf

tests/doubles/woocommerce-seo-double.php
```diff
@@ -20,4 +20,11 @@ class Yoast_WooCommerce_SEO_Double extends Yoast_WooCommerce_SEO {
 	public function check_dependencies( $wp_version ) {
 		return parent::check_dependencies( $wp_version );
 	}
+
+	/**
+	 * @inheritdoc
+	 */
```
dutchmartin: This seems not the right way to use this tag : https://docs.phpdoc.org/guides/inheritance.html
Travis fails on this too, but when the correct tag is used, it should be fixed: https://github.com/squizlabs/PHP_CodeSniffer/pull/214

dutchmartin:
There are merge conflicts on this PR, can you please resolve them.

dutchmartin:
acceptance done 👍

PR #260, opened 2018-06-19
Excludes the WooCommerce core pages from the sitemap

andizer:
## Summary

* Excludes the WooCommerce core pages from the sitemap, because they will set to noindex by WooCommerce.

## Test instructions

* Set a myaccount, cart and checkout page.
* Visit the pages sitemap and see the set pages not being visible.

Fixes https://github.com/Yoast/wordpress-seo/issues/9540

### DIFF
--- a/wpseo-woocommerce.php
+++ b/wpseo-woocommerce.php
@@ -226,6 +226,25 @@ class Yoast_WooCommerce_SEO {
 		if ( $this->license_manager ) {
 			add_action( 'admin_init', array( $this, 'init_beacon' ) );
 		}
+
+		add_filter( 'wpseo_exclude_from_sitemap_by_post_ids', array( $this, 'filter_woocommerce_pages' ) );
+	}
+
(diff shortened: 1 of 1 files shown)


wpseo-woocommerce.php
```diff
@@ -226,6 +226,25 @@ protected function initialize() {
 		if ( $this->license_manager ) {
 			add_action( 'admin_init', array( $this, 'init_beacon' ) );
 		}
+
+		add_filter( 'wpseo_exclude_from_sitemap_by_post_ids', array( $this, 'filter_woocommerce_pages' ) );
+	}
+
+	/**
+	 * Adds the page ids fro the WooCommerce core pages to the excluded post ids.
```
dutchmartin: I would suggest using "Adds the page **id's from** the WooCommerce core pages to the excluded post id's"

dutchmartin:
Acceptance done

PR #309, opened 2018-12-18
Update for YoastCS 1.1.0 / WPCS 1.2.1 / PHPCompatibility ^9.0.0

jrfnl:
### Composer: use YoastCS 1.1.0+

* Includes adding the PHPCS XSD to the custom ruleset which should now be valid as the minimum PHPCS requirement has gone up to PHPCS 3.3.2.
* Includes removing the DealerDirect Composer PHPCS plugin which now comes automatically with YoastCS.

### PHPCS: remove redundant ignore annotation

The typical issues reported for `index.php` files are ignored by default in YoastCS.

### CS/QA: deprecated functions consistency

This PR addresses some inconsistencies related to deprecated functions. It basically applies the following rules:
* Deprecated functions/methods should be at the end of the file/class.
* They should be preceded by a `/* *** DEPRECATED METHODS *** */` delimiter to clearly indicate the start of the deprecated functions.
* Deprecated functions should be ignored for the purposes of calculating code coverage via PHPUnit and therefore should each have a `@codeCoverageIgnore` tag in their docblock.

### DIFF
--- a/.phpcs.xml.dist
+++ b/.phpcs.xml.dist
@@ -1,5 +1,8 @@
 <?xml version="1.0"?>
-<ruleset name="Yoast WooCommerce SEO">
+<ruleset xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	name="Yoast WooCommerce SEO"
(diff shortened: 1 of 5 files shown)


andizer:
CR done 👍

PR #316, opened 2019-01-22
PHPCS ruleset: update the prefixes array

jrfnl:
## Summary
* _N/A_

Updates the PHPCS prefixes array to cover the new prefixes which should be used from now on for global structures as well as namespaced structures based on the agreed prefixes per September 2018.

## Test instructions

* _N/A_

The change can be tested, but this isn't really needed as you'd be testing PHPCS + WPCS (which test this anyway).


andizer:
CR done 👍

PR #342, opened 2019-04-09
Fix woo schema

jdevalk:
## Summary

* Use the WooCommerce Schema output and stitch it into the Yoast SEO Schema graph.

* Decided to block Woo from outputting its own schema, but still use its data collection features.
* Brings back `brand` and `manufacturer` attributes that I missed in the previous merge.

## Test instructions

* Look at a product page for a product with multiple reviews, throw the output into the SDTT, and it should stitch into one graph.


moorscode:
CR 👍 
Quick acceptance test 👍

PR #363, opened 2019-05-07
QA: use the most specific PHPUnit assertion possible

jrfnl:
## Summary
* _N/A_

This is a long established best practice.

PHPUnit contains a variety of assertions and the ones available have been extended hugely over the years.
To have the most reliable tests, the most specific assertion should be used.

This implements this for the Yoast WooCommerce plugin.

Refs:
* https://phpunit.de/manual/4.8/en/appendixes.assertions.html
* https://phpunit.readthedocs.io/en/7.5/assertions.html#

Most notably, this changes calls to `assertEquals()` to `assertSame()`, where `assertEquals()` does a loose type comparison and `assertSame()` does a strict type comparison.

Refs:
* https://phpunit.de/manual/6.5/en/appendixes.assertions.html#appendixes.assertions.assertEquals
* https://phpunit.de/manual/6.5/en/appendixes.assertions.html#appendixes.assertions.assertSame

Note that this change has not been made for the `WPSEO_WooCommerce_Beacon_Setting_Test::test_get_products_for_the_woocommerce_seo_page()` test as two different instances of the same class will not be the _same_, though they can be _equal_.

## Test instructions

* As long as the unit tests pass in the Travis builds, we are good.

### DIFF
--- a/tests/classes/option-woo-test.php
+++ b/tests/classes/option-woo-test.php
@@ -33,7 +33,7 @@ class WPSEO_Option_Woo_Test extends WPSEO_WooCommerce_UnitTestCase {
 	 */
 	public function test_constructor() {
 		$option = new WPSEO_Option_Woo_Double();
-		$this->assertEquals(
+		$this->assertSame(
 			array(
 				'price' => 'Price',
 				'stock' => 'Stock',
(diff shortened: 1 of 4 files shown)


IreneStr:
CR 👍

PR #364, opened 2019-05-07
Composer/PHPUnit: be explicit about supported PHPUnit versions

jrfnl:
## Summary
* _N/A_

The unit tests for the Yoast WooCommerce plugin, in part, depend on the WordPress native unit test setup and loads the WP native `bootstrap.php` file to that end.

This means that the unit tests for the plugin are compatible with the same PHPUnit versions as the WP native unit tests, which at this time is `^3.6.12 || ^4.5 || ^5.7 || ^6.0 || ^7.0`.
While PHPUnit 3.x is supported for testing on PHP 5.2, it is not listed in the `composer.json` file as Composer itself has a minimum requirement of PHP 5.3, so listing it is a moot point.

As PHPUnit is now included in the Composer setup, the Travis script needs to be tweaked:
* Defer to the Travis native PHPUnit version when it is within our range for faster builds.
    All Travis PHP images ship with PHPUnit already installed.
    It's only for those images which ship with an incompatible PHPUnit version (PHP 7.2 and higher ship with PHPUnit 8.x), that we need to install and use the PHPUnit version declared in the `composer.json` file.
* Make the `composer install` more efficient by only installing what we actually need and removing what we don't need _before_ the install is done.

### CI/QA: make the phpunit.xml.dist file cross-version compatible

The specs for the `phpunit.xml.dist` file have changed numerous times over the years with new options being added and outdated options being removed.
These specs are contained in a schema XSD file for each minor PHPUnit version.

This commit:
* Adds a proper XML file header;
* Annotates that an XSD schema is available and used;
* Adds a `name` to the testsuite.
    As of PHPUnit 7, each testsuite is required to have a `name` attribute. Prior to that version, the `name` attribute was already allowed, so adding the attribute improves cross-version compatibility of the config file.
* Removes the top-level `strict` directive which [has been removed in PHPUnit 5](https://phpunit.de/announcements/phpunit-5.html). It was already deprecated before that time.
* Removes the top-level `syntaxCheck` directive which was removed prior to PHPUnit 3.
* Fixes the indentation to tabs.

What I have not changed:
* The `testsuite` node was not allowed according to the XSD until PHPUnit 4.4, however as the config with the `testsuite` works just fine in PHPUnit 3.x and 4.x, this is not an issue.

## Test instructions

* Check the output of the Travis builds to see that the unit tests run in all builds and use the correct PHPUnit versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,15 +74,16 @@ before_install:
 
 install:
 - if [[ $TRAVIS_PHP_VERSION == "5.2" ]]; then phpenv local 5.6.13; fi
+- if [[ $TRAVIS_PHP_VERSION == "5.2" ]]; then composer remove --dev --no-update --no-scripts yoast/yoastcs dealerdirect/phpcodesniffer-composer-installer; fi
 - |
   if [[ "$PHPCS" == "1" ]]; then
     composer install --no-interaction
     composer config-yoastcs
+  elif [[ ${TRAVIS_PHP_VERSION:0:3} > "7.1" ]]; then
+    composer install --no-interaction
   else
     composer install --no-dev --no-interaction
   fi
-- if [[ $TRAVIS_PHP_VERSION == "5.2" ]]; then composer remove --dev yoast/yoastcs dealerdirect/phpcodesniffer-composer-installer; fi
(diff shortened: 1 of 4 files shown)


IreneStr:
CR 👍

PR #371, opened 2019-05-14
QA: review test bootstrap file / show friendly errors

jrfnl:
## Summary

* _N/A_

### Test bootstrap: remove unnecessary require

WP will load the active plugins automatically, there is no need to require YoastSEO explicitly.

### Test bootstrap: show friendly error message on failure

Do more rigorous checking whether the environment is correctly set up for the unit tests to be run and if not, show an instructive error message.

Include disabling the security sniffs for the bootstrap file.

## Test instructions

On `trunk`:
1. If applicable: remove the `WP_DEVELOP_DIR` environment variable from your OS or custom `phpunit.xml` file and make sure that this plugin is not in the `wp-content/plugins` directory of a WP install. Run the unit tests and see them fail with a PHP error.
2. Bring back the `WP_DEVELOP_DIR` environment variable, but set it to an invalid value. Run the unit tests and see them fail with a PHP error.
3. Revert the change(s) made in [1] and [2].
4. Remove the `WP_PLUGIN_DIR` environment variable from your OS or custom `phpunit.xml` file and make sure that this plugin is not in the `wp-content/plugins` directory of a WP install. Run the unit tests and see them fail with a PHP error.
5. Revert the change(s) made in [4].
6. Rename the `vendor` directory to `vendorx`. Run the unit tests and see them fail with a PHP error.
7. Revert the change(s) made in [6].

Now check out this branch, follow the same steps as above, but instead of PHP errors, you should see instructive error message for each of the above situations.

### DIFF
--- a/.phpcs.xml.dist
+++ b/.phpcs.xml.dist
@@ -122,8 +122,9 @@
 		<exclude-pattern>/tests/doubles/*\.php$</exclude-pattern>
 	</rule>
 
-	<!-- Allow for the unit test case file to make changes to the superglobals. -->
+	<!-- Allow for select unit test files to use less secure code. -->
 	<rule ref="WordPress.Security">
+		<exclude-pattern>/tests/bootstrap\.php$</exclude-pattern>
 		<exclude-pattern>/tests/framework/woocommerce-unittestcase\.php$</exclude-pattern>
 	</rule>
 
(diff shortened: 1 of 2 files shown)


Xyfi:
Acceptance 🚧 

I got most of the described error messages except for `ERROR: Run composer install to generate the autoload files before running the unit tests.` when renaming the vendor folder.

PR #390, opened 2019-07-18
Don't add image id if no thumbnail is available

Xyfi:
## Summary

* Don't output image schema id if the product has no thumbnail.

* Removes image schema if it is false. This is a bug from WooCommerce.
* If a product has no defined image, fallback to WooCommerce placeholder image.

## Test instructions

Enable `WooCommerce`, `Yoast SEO: WooCommerce` and `Yoast SEO`.

* Create a product without a product image.
* Inspect the product schema, it should not have an image node.
* Add a product image.
* Inspect the product schema, it should now have an `image` node with an `@id` property referring to the main image schema.

Fixes Yoast/bugreports#458

### DIFF
--- a/classes/woocommerce-schema.php
+++ b/classes/woocommerce-schema.php
@@ -193,7 +193,12 @@ class WPSEO_WooCommerce_Schema {
 	 * @param string $canonical The product canonical.
 	 */
 	private function add_image( $canonical ) {
-		// WooCommerce will set the image to false if non is available. This is incorrect schema.
+		/**
+		 * WooCommerce will set the image to false if non is available. This is incorrect schema and we should fix it
+		 * for our users for now.
+		 *
+		 * See https://github.com/woocommerce/woocommerce/issues/24188.
+		 */
 		if ( $this->data['image'] === false ) {
 			unset( $this->data['image'] );
 		}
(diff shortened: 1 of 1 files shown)


IreneStr:
CR 👍

IreneStr:
Acceptance 👎 

The specs say:
> ACTION: When an image isn't set, we should generate a node for the placeholder image, and reference that instead (if it exists and is a valid imageObject).

However, this PR removes the image node altogether.

classes/woocommerce-schema.php
```diff
@@ -189,4 +186,33 @@ private function add_organization_for_attribute( $attribute, $product, $taxonomy
...
+		}
+
+		if ( has_post_thumbnail() ) {
+			$this->data['image'] = array(
+				'@id' => $canonical . WPSEO_Schema_IDs::PRIMARY_IMAGE_HASH,
+			);
+		}
+
+		// Fallback to WooCommerce placeholder image.
+		if ( function_exists( 'wc_placeholder_img_src' ) ) {
```
IreneStr: Because you don't return in the previous code block, and `wc_placeholder_img_src` always exists, the woocommerceimageplaceholder will always overwrite the `$this->data['image']`.

afercia:
CR 👍 
Added missing `return` and adjusted a couple minor typos.

classes/woocommerce-schema.php
IreneStr: Fixed by @afercia in d2d21ec1b817f39be1edbf92ef00398c3675db7e

IreneStr:
Acceptance 👍

PR #396, opened 2019-08-01
Composer: require YoastCS ^1.3.0 & update ruleset

jrfnl:
## Summary

*  _N/A_

Ruleset updates:
* Remove dependencies related `exclude-pattern`. These are now contained within the YoastCS ruleset as of v1.3.0.
* Remove the `languages` directory `exclude-pattern`.
    The languages directory contains no PHP files, except for the empty `index.php` file, so there is no need to exclude it.
    Includes very minor touch-up of the `index.php` file.
* Remove the `minimum_supported_wp_version` `config` directive. As of YoastCS 1.3.0, a default is set in the `Yoast` ruleset and should be followed by this repo.
    The default in the `Yoast` ruleset at this time is `4..9`.
* Move the property setting for the `custom_test_class_whitelist` property which is used by several sniffs up to the standard inclusion.
    This allows us to get rid of the list duplication.
* Minor spelling fix in inline documentation.

Includes two minor whitespace fixes to the actual code files to comply with YoastCS 1.3.0.

## Test instructions

* _N/A_. If the build passes, we're good.

### DIFF
--- a/.phpcs.xml.dist
+++ b/.phpcs.xml.dist
@@ -14,10 +14,6 @@
 
 	<file>.</file>
 
-	<exclude-pattern>vendor/*</exclude-pattern>
-	<exclude-pattern>node_modules/*</exclude-pattern>
-	<exclude-pattern>languages/*</exclude-pattern>
-
 	<!-- Only check PHP files. -->
 	<arg name="extensions" value="php"/>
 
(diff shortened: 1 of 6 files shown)


igorschoester:
@jrfnl Rebase required

jrfnl:
@igorschoester Done!

PR #402, opened 2019-08-13
Add primary term check to the schema taxonomy

igorschoester:
## Summary

* Fixes a bug where the primary category would not be reflected correctly in the product schema output when it was being used to set the manufacturer or brand.

* Added tests for general schema output to be able to test the private functions that are changed.
* Changed the class variables from private to protected to be able to set them in the test double.
* Moved the `Frontend->canonical` to a protected function to be able to mock it.
* Added PHPCS exclusions for the test doubles.

## Test instructions

* Reproduce on trunk: see the issue.
* Switch to this branch and build it.
* Refresh the frontend source of your product again and check the output of the `brand` and `manufacturer` to see if they match the primary category.

Fixes https://github.com/Yoast/bugreports/issues/468


classes/woocommerce-schema.php
```diff
@@ -217,4 +215,45 @@ private function add_image( $canonical ) {
...
+	 * @param int    $post_id       Post ID for the term.
+	 *
+	 * @return WP_Term|null The primary term, the first term or null.
+	 */
+	protected function get_primary_term_or_first_term( $taxonomy_name, $post_id ) {
+		$primary_term    = new WPSEO_Primary_Term( $taxonomy_name, $post_id );
+		$primary_term_id = $primary_term->get_primary_term();
+
+		if ( $primary_term_id !== false ) {
+			$primary_term = get_term( $primary_term_id );
```
IreneStr: Reusing the $primary_term variable in this scope is a little confusing.

classes/woocommerce-schema.php
```diff
@@ -217,4 +215,45 @@ private function add_image( $canonical ) {
 			$this->data['image'] = $image_schema->generate_from_url( wc_placeholder_img_src() );
 		}
 	}
+
+	/**
+	 * Tries to get the primary term, then the first term, null if none found.
+	 *
+	 * @codeCoverageIgnore
```
IreneStr: I don't think there is a reason to not test this function.

IreneStr:
Acceptance 👍

IreneStr:
CR of new tests 👍

PR #405, opened 2019-08-22
Improve grunt artifact

andizer:
This pull requests implements the same logic as used for the wpseo-news plugin. It also removes some packages because these are already included in `@yoast/grunt-plugin-tasks`.

A zip file can be build with `grunt artifact`. I've compared the result of this command by checking out the 11.9 tag and running the command. After this I compared the content of the zip with the content of the latest released version (11.9). 

I've done the comparison by using colordiff. `brew install colordiff` and then just do enter the next command in the commandline: `colordiff -r dirname1 dirname2`


moorscode:
I think many configs; that already exist in the `plugin-grunt-tasks` repository can be removed from this repo if they are the same. Only the exceptions should remain locally to override the default behaviour.

Otherwise I think it looks good - from a code perspective 👍

Xyfi:
After looking at this PR during our sprint refinement meeting we have the following comments:

Before we can pick this up we would like @moorscode's comment to be resolved and we'dd like to see proper test instructions (using `colordiff` for example).

Gruntfile.js
```diff
@@ -45,16 +63,16 @@ module.exports = function( grunt ) {
...
 	loadGruntConfig( grunt, {
-		configPath: path.join( process.cwd(), project.paths.config ),
+		configPath: path.join( process.cwd(), "node_modules/@yoast/grunt-plugin-tasks/config/" ),
+		overridePath: path.join( process.cwd(), project.paths.config ),
 		data: project,
 		jitGrunt: {
 			staticMappings: {
 				addtextdomain: "grunt-wp-i18n",
 				makepot: "grunt-wp-i18n",
 				glotpress_download: "grunt-glotpress",
```
IreneStr:
```suggestion
				glotpress_download: "grunt-glotpress",
				wpcss: "grunt-wp-css",
```

IreneStr:
CR 👍

IreneStr:
Acceptance 👍:
- the built artifact folders are the same (except for the autoloader hash and the pot file creation time).
- grunt set-version, update-version, eslint, check, build:i18n, and phpcs all work correctly as well.

PR #409, opened 2019-09-12
PHPCS: Revert exclusion of test directory

### LINKED ISSUE #408: Fix CS issues for the tests
I've excluded the tests directory (See https://github.com/Yoast/wpseo-woocommerce/pull/405/files#diff-325a505e494aa8078473d0b71ad82200R20) from being checked for the PHP CS, this is because the 'plugin-grunt-tasks' dependency has been updated causing a lot of new CS errors. Because of fixing these errors were out of scope I decided to exclude the tests directory.

When we stop supporting PHP 5.2 we can remove the exclusion and maybe then fix possible CS errors.

jrfnl:
## Summary

* _N/A_

* Revert the exclusion of the `tests` directory from the CS run.
* Use the right command in `grunt` to call PHPCS.

The Composer script was already correctly set up to allow for different minimum PHP requirements for the `tests` directory, it just needs to be used. 

## Test instructions

@andizer Got a feeling you already know how to test this.

Fixes #408

### DIFF
--- a/.phpcs.xml.dist
+++ b/.phpcs.xml.dist
@@ -17,7 +17,6 @@
 	<exclude-pattern>vendor/*</exclude-pattern>
 	<exclude-pattern>node_modules/*</exclude-pattern>
 	<exclude-pattern>languages/*</exclude-pattern>
-	<exclude-pattern>tests/*</exclude-pattern>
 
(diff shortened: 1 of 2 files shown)


igorschoester:
CR & Acceptance ✅

PR #413, opened 2019-10-02
CS: enforce consistent array format

jrfnl:
## Summary
* _N/A_

WordPressCS will demand long arrays and forbid the use of short arrays as of WPCS 2.2.0.
See: https://make.wordpress.org/core/2019/07/12/php-coding-standards-changes/

For YoastCS, however, a choice has been made to demand short arrays.

For now, while WP 4.9, and therefore PHP 5.2 still needs to be supported, this is only possible in the directories containing files with PHP 5.6+ code.
So in this interim period, we will enforce short arrays in the PHP 5.6+ code and long arrays everywhere else.

Once the minimum supported WP version has moved up to WP 5.2 / PHP 5.6, short arrays will be enforced for all code and the change-over can be done automatically using the auto-fixer included in the relevant sniff.

## Test instructions

* _N/A_
    This is a code-only change and should have no effect on the functionality.

### DIFF
--- a/.phpcs.xml.dist
+++ b/.phpcs.xml.dist
@@ -130,4 +130,21 @@
 		<exclude-pattern>/tests/TestCase\.php$</exclude-pattern>
 	</rule>
 
+
+	<!--
+	#############################################################################
+	TEMPORARY TWEAK
+	YoastCS will demand short arrays, but until support for WP < 5.2 has been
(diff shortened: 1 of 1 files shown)


johannadevos:
> plugin doesn't currently _have_ a `src` directory...

@jrfnl Ah right, sorry

johannadevos:
CR: 👍

PR #433, opened 2019-11-01
Stop explicit support for PHP < 5.6 / WP < 5.2

jrfnl:
## Summary
* _N/A_

Brief analysis of the Travis script shows me that:
* There are a number of builds _which do not do anything_.
    All the tasks in `script` are set up conditionally based on environment variables.
    Any build which did not have an environment variable, such as `PHPLINT`, `PHPCS` or `PHPUNIT` was basically completely useless.
* While the `PHPUNIT` task is set up to run conditionally, the _preparation of the environment_ for unit testing, i.e. downloading WordPress, cloning YoastSEO Free, setting up the database _is all done **unconditionally**_.

So, with that in mind, I have made the following changes to the Travis script - please read this carefully:
* Removing builds against PHP < 5.6.
* Changing the `WP_VERSION` environment variables in the matrix to use a minimum of `5.2`.
    Note: this will change again once WP 5.3 comes out.
* Actually running the unit tests on all current builds.
* Removing the PHP version/env condition toggle for a full/no-dev composer install.
    As a composer install was being run on all builds anyway in one way or another **and** there is a committed `composer.lock` file **and** Travis caching of the downloads is setup, doing a full install each time will not make the build significantly slower.
    Also: It is better to always test based on the dependencies we've set up than to rely on  a PHP image created by Travis which may fail unexpectedly (especially as Travis has seen some issues recently with incorrect PHPUnit versions on certain images).
    This also means the `composer dump-autoload` can be removed as it will now always have run whenever the autoload is needed.
    **Note**: this changes the `composer install` for the `deploy` stage from a full install to a `no-dev` install.
   That stage was currently being run against PHP 7.2, so was getting a full install, while AFAICS that shouldn't be needed.
* Making the environment preparations which are needed for the unit tests dependent on the `PHPUNIT` environment variable.
* Making the builds slightly faster by using lean git clones (`depth=1` instead of `depth=50`).
* Using the composer scripts to run the unit test tasks.
    As PHPUnit will now always be available in `vendor`, using the composer scripts simplifies things.

Includes:
* Adjusting the `Requires...` info in the `readme.md` file.
* Removing the `xrstf/composer-php52` dependency and references to the `vendor/autoload_52.php` file.
* Dropping PHPUnit < 5.7 from the Composer `require-dev`.

Also:
* Travis hasn't supported `sudo` for quite a while now, so removing it.

Related #425

Related to Yoast/wordpress-seo#13758

## Test instructions

* Check the detailed output of the Travis scripts to verify that all is working as expected.

**_Please be extra critical reviewing this PR for any changes which may impact the `deploy` stage as those won't show up during the PR build._**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: php
 dist: trusty
-sudo: false
 
 branches:
   only:
@@ -16,23 +15,15 @@ jobs:
   fast_finish: true
   include:
     - php: 7.3
-      env: WP_VERSION=5.0 WP_MULTISITE=1 PHPLINT=1 PHPCS=1 SECURITY=1
-    - php: 5.2
-      # As 'trusty' is not supporting PHP 5.2/5.3 anymore, we need to force using 'precise'.
-      dist: precise
-      env: WP_VERSION=4.8 WP_MULTISITE=1 PHPLINT=1 PHPUNIT=1
+      env: WP_VERSION=5.2 WP_MULTISITE=1 PHPLINT=1 PHPUNIT=1 PHPCS=1 SECURITY=1
     - php: 5.6
-      env: WP_VERSION=5.0 PHPUNIT=1
+      env: WP_VERSION=5.2 PHPLINT=1 PHPUNIT=1
     - php: 7.0
-      env: WP_VERSION=4.9
-    - php: 5.2
-      # As 'trusty' is not supporting PHP 5.2/5.3 anymore, we need to force using 'precise'.
-      dist: precise
-      env: WP_VERSION=master
+      env: WP_VERSION=master WP_MULTISITE=1 PHPUNIT=1
     - php: "7.4snapshot"
-      env: WP_VERSION=master
+      env: WP_VERSION=master PHPUNIT=1
     - stage: deploy
-      env: WP_VERSION=4.9
+      env: WP_VERSION=5.2
       if: tag IS present
       before_install:
         - openssl aes-256-cbc -K $encrypted_2b922af4d08d_key -iv $encrypted_2b922af4d08d_iv -in ./deploy_keys/wpseo_woo_deploy.enc -out ./deploy_keys/wpseo_woo_deploy -d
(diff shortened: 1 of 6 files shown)


andizer:
Code looks good to me. What is the basis to have a deploy check? This one didn't exists before? I have no clue how to test that one?

jrfnl:
> 
> 
> Code looks good to me. What is the basis to have a deploy check? This one didn't exists before? I have no clue how to test that one?

You mean like this ?

https://github.com/Yoast/wpseo-woocommerce/pull/433/files#diff-354f30a63fb0907d4ad57269548329e3R70

The `composer install` previously had a PHP version based check - `${TRAVIS_PHP_VERSION:0:3} > "7.1"` - and the `deploy` stage silently and implicitly relied on that.

What I mean by that is that the `deploy` stage does not have a PHP version set and thus gets the Travis default which may change at any time.
As the default is currently PHP `7.2`, the PHP version based condition was triggered for `deploy` without it being explicit in the script that the `deploy` stage _needs_ the `composer install` - at least, I presume it needs it.

Either way, as the PHP version based check is now replaced by a ENV keyword check and the `deploy` stage doesn't have any of the relevant keywords, the change on the line mentioned above, makes it:
- explicit that `deploy` needs a `composer install`;
- maintains the same behaviour as previously for the `deploy` stage.

Does that explain it well enough ?

andizer:
There is a merge conflict, can you rebase this pull?

dutchmartin:
@jrfnl I do not work for Yoast anymore. For hiring me as a consultant, send me a mail at  `info@martijngroeneveldt.nl`.

jrfnl:
@andizer Rebased & force pushed.

jrfnl:
> 
> 
> @jrfnl I do not work for Yoast anymore. For hiring me as a consultant, send me a mail at `info@martijngroeneveldt.nl`.

I appreciate the self-promo, but that's neither here nor there. Good luck in your endeavors.

PR #435, opened 2019-11-06
Remove all hidden products from XML sitemap

jdevalk:
## Summary

* Fixes a bug where hidden products could show in XML sitemap.

* Used an unbounded query, as it only grabs the IDs, not more data, I think this won't lead to any problems.

## Test instructions

* See reproduction steps in https://github.com/Yoast/bugreports/issues/697

Props to @Djennez for the solution.

Fixes https://github.com/Yoast/bugreports/issues/697


igorschoester:
CR & Acceptance ✅

PR #436, opened 2019-11-11
Retrieves noindex from shop page instead of product archive

### LINKED ISSUE #401: WooCommerce Shop Base page in Products XML sitemap
Hi,

In WooCommerce I have chosen to use the homepage as the shoppage.

When I look in my products xml sitemap, the shoppage is listed there as well, see below image, this is no good because that URL should not be indexed at all!

What to do?

Kind regards,

Willem

stodorovic:
## Summary

* Make sure that when shop page is noindexed it doesn't show in XML sitemap.

## Technical decisions

* Uses the filter `wpseo_sitemap_page_for_post_type_archive` to get noindex from shop page instead of product archive.

## Test instructions

*  Install [WooCommerce](https://wordpress.org/plugins/woocommerce/), create and set "shop" page.
* Check both sitemaps (page and product). This page should be listed in both sitemaps.
* Set _noindex_ for "shop" page.
* Check both sitemaps (page and product). This page should be removed in both sitemaps.

Partially fixes #401


jdevalk:
CR and Acceptance: 👍

PR #446, opened 2019-12-01
Composer: update BrainMonkey dependency

jrfnl:
## Summary
* _N/A_

BrainMonkey has released a new version.

Significant changes which are reasons to upgrade:
* Compatibility with PHP 7.4.
* Build in support for calls to `do_action_deprecated()` and `apply_filters_deprecated()`.

Note: the `composer update` has been run in a way so as to _only_ update BrainMonkey and the associated dependencies. Nothing else.

Refs:
* https://github.com/Brain-WP/BrainMonkey/releases/tag/2.4.0

## Test instructions

* _N/A_
    This is a test dependency-only change and should have no effect on the functionality. If the Travis build still passes, we're good.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
     "require-dev": {
         "yoast/yoastcs": "^1.3.0",
         "phpunit/phpunit": "^5.7 || ^6.0 || ^7.0",
-        "brain/monkey": "^2.2"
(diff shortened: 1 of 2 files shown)


johannadevos:
CR: 👍

PR #455, opened 2020-01-03
Filters the retrieved taxonomies to only have the viewable ones.

### LINKED ISSUE #454: Hide hidden taxonomies in the settings
In WooCommerce we can choose the fields to show as manufacturer and brand. But the dropdown also shows non public taxonomies that are related to WooCommerce. 

Cursor_and_WooCommerce_SEO_-_Yoast_SEO_‹_Local_WordPress_Dev_—_WordPress

The two tags are the product_type and product_variation. And our prominent words is visible.

### Todo
- [ ] Check if the taxonomy is public accessible here https://github.com/Yoast/wpseo-woocommerce/blob/trunk/wpseo-woocommerce.php#L480 - We probably can use the function [is_taxonomy_viewable](https://developer.wordpress.org/reference/functions/is_taxonomy_viewable) for this.

andizer:
## Summary

* Fixes a bug where non-public taxonomies can chosen as manufacturer and brand.

*

## Test instructions

* Checkout trunk
* In the admin settings you will see prominent words (when premium enabled) and two times tags.
* Checkout this branch
* Now you only see the public taxonomies.

Fixes #454

### DIFF
--- a/wpseo-woocommerce.php
+++ b/wpseo-woocommerce.php
@@ -474,7 +474,7 @@ class Yoast_WooCommerce_SEO {
 	public function admin_panel() {
 		Yoast_Form::get_instance()->admin_header( true, 'wpseo_woo' );
 
-		$object_taxonomies = get_object_taxonomies( 'product', 'objects' );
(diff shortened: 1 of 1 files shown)


jdevalk:
Tested and approved.

PR #457, opened 2020-01-07
Adds SKU to the output, fixes instock to be in stock

### LINKED ISSUE #456: [Feature Request] Enhancements to product meta tags for Facebook Catalog
Facebook Catalog (the product feed system which powers Facebook Marketplace) is able to extract open graph / schema data from a page in order to automatically populate or update a store's inventory.

Our current meta tags require only a few tweaks in order to make them compatible with this feature.

## Requirements

We should do the following on product templates:
- Add a `<meta property="product:retailer_item_id" content="{{SKU}}" />` tag, where `{{SKU}}` is the product SKU.
- Add a `<meta property="product:brand" content="{{BRAND_NAME}}" />` tag, where `{{BRAND_NAME}}`.
- Alter the value of the `product:availability` tag from `instock` to `in stock`.

## Future considerations
We should also support a `<meta property="product:category" content="{{CATEGORY}}" />` tag. However, this will require some UI work - the tag value must be one of the values [from here](https://www.google.com/basepages/producttype/taxonomy.en-US.txt), which would necessitate a selector interface.

That's outside the scope of this ticket, but worth being aware of.

jdevalk:
## Summary

* Adds `product:retailer_item_id` to the plugins OpenGraph output for Facebook Catalog usage.
* Fixes a bug where `product:availability` was set to `instock` instead of the correct `in stock`.

## Test instructions

* Go to a product page, see the output now contains `product:retailer_item_id` and `product:availability` is set correctly.

Fixes #456

### DIFF
--- a/wpseo-woocommerce.php
+++ b/wpseo-woocommerce.php
@@ -752,7 +752,7 @@ class Yoast_WooCommerce_SEO {
 			echo '<meta property="product:availability" content="in stock" />' . "\n";
 		}
 
-		echo '<meta property="product:retailer_item_id" content="' . $product->get_sku() . '" />' . "\n";
+		echo '<meta property="product:retailer_item_id" content="' . esc_attr( $product->get_sku() ) . '" />' . "\n";
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


jcomack:
Acceptance done 👍

PR #463, opened 2020-01-07
If the shop page isn't set in Woo, don't try to use it in the XML sitemap

### LINKED ISSUE #401: WooCommerce Shop Base page in Products XML sitemap

jdevalk:
## Summary

* When the shop page isn't set in WooCommerce, don't try to use it in the XML sitemap.

Fixes #401


jcomack:
Acceptance done 👍

PR #464, opened 2020-01-07
When Organization isn't set, don't use it in our Schema output.

### LINKED ISSUE #399: Products shouldn't reference a seller if no organization is set
Product offers have a `seller` attribute, which is meant to reference the `Organization` 'behind' the website. However, if the `Organization` is not set/present (e.g., due to validation issues), the ID reference remains regardless, despite not referencing a valid piece.

image

Nodes shouldn't reference entities which don't exist.

Specifically, in this case, the `seller` property should check that the `organization` it's referencing exists within the graph.

jdevalk:
## Summary

* When Organization isn't set, don't use it in our Schema output.

## Test instructions

* See #399

Fixes #399

### DIFF
--- a/classes/woocommerce-schema.php
+++ b/classes/woocommerce-schema.php
@@ -98,14 +98,7 @@ class WPSEO_WooCommerce_Schema {
 	public function change_product( $data, $product ) {
 		$canonical = $this->get_canonical();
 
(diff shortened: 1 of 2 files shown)


jcomack:
CR done 👍

jcomack:
Acceptance done 👍

PR #472, opened 2020-01-09
Remove distribution files from repository

### LINKED ISSUE #240: Remove the distribution files (JS/CSS) from the repository
Like we are doing on [Wordpress SEO](https://github.com/Yoast/wordpress-seo/issues/9119), we should remove the distribution files (JS/CSS) from the repository since travis will build these when needed.

Another thing to do in this issue is moving the location of the distribution files when they get build.
For example: there are still distribution files in the `/js` directory, they should be moved to a `/js/dist `folder so the file structure is easy to understand.

jdevalk:
## Summary

* None needed.

*  Moved the JS files to a `dist` directory within the `js` directory.
* Updated plugin-grunt-tasks to the last minor version in the same go.

## Test instructions

* Make sure the right files are still loaded on product edit screens.

Fixes #240


jcomack:
Acceptance done 👍

PR #487, opened 2020-01-21
When a product has an ISBN, it's Schema technically a Book

jdevalk:
## Summary

* When adding an ISBN number to a product, we double "type" it to `Book / Product`. By double typing it, the Product can have the attributes of both [schema.org/Book](https://schema.org/Book) _and_ [schema.org/Product](https://schema.org/Product), and thus it can have an `isbn` attribute and a `price` etc.

## Test instructions

* Add an ISBN number to a product, see that the Schema output now contains `[ 'Book', 'Product' ]` as `@type`.


jcomack:
Acceptance done 👍

PR #488, opened 2020-01-21
Don't exclude product attributes from the XML sitemap by default

jdevalk:
## Summary

* Currently, no product attributes are ever added to the Yoast SEO XML sitemap. With this fix, attributes which have archives _are_ added to the XML sitemaps.

## Test instructions

* Create two product attributes. Set one to have archives, the other not to have archives. See that the former shows up in the XML sitemap while the latter doesn't.


jcomack:
Acceptance done 👍

PR #493, opened 2020-01-27
Make the versioning setup the same as news seo

### LINKED ISSUE #491: WooCommerce 12.5 generates a console log error
While testing Yoast SEO: WooCommerce 12.5-RC2 I've run into an error on the Products page for WooCommerce. The error mentions our previous version, 12.4.1.

### How can we reproduce this behaviour?
1. With WooCommerce 3.9.0, Yoast SEO 13.0-RC1, and Yoast SEO: WooCommerce 12.5-RC2 active
2. When you navigate to the WooCommerce Products page
3. You'll see this error in the console log:
Schermafbeelding 2020-01-22 om 14 52 02

andizer:
## Summary

* N/A - Tooling - The version update command didn't work.

* Decided to make an extra constant that feeds the value of class constant `Yoast_WooCommerce_SEO::version` - We should remove the last one when the time is there.
* I also removed a faulty config that relies on an old location. I guess the class has been moved to another file in the meanwhile.

## Test instructions

* Set a new version and run the update-version command
* The version should be updated now.

Fixes #491


hansjovis:
Acceptance test: ✅ 

> I also removed a faulty config that relies on an old location. I guess the class has been moved to another file in the meanwhile.

By removing the faulty config, we use the default configuration of the `set-version` task. The default config is [specified here](https://github.com/Yoast/plugin-grunt-tasks/blob/develop/config/update-version.js).

The default configuration searches for a string in the form of 
```
define( '<%= pluginVersionConstant %>', 'version number' );
```

PR #507, opened 2020-01-29
Make replace breadcrumbs work with storefront

### LINKED ISSUE #502: WooCommerce breadcrumbs not replaced
### What did you expect to happen?
After enabling the WPSEO WooCommerce option to replace the WooCommerce breadcrumbs with the Yoast ones, for that to work.

### What happened instead?
The WC breadcrumbs were not replaced.

### How can we reproduce this behavior?
1. Set up a store with the needed plugins (see below).
2. Use the Storefront theme (as that will have WC breadcrumbs)
3. Enable `Replace WooCommerce Breadcrumbs` in the WPSEO WC options (`SEO -> WooCommerce SEO` )
4. Refresh a page with a breadcrumb and see it not change

### Technical info
* WordPress version: 5.3.2
* Yoast SEO version: 13.0-RC1
* WooCommerce SEO version: 12.5-RC3

At the moment where we check if we should replace the breadcrumbs, the needed options are not loaded in the Yoast options class. So the check 

```php
		if ( WPSEO_Options::get( 'woo_breadcrumbs' ) !== true || WPSEO_Options::get( 'breadcrumbs-enable' ) !== true ) {
			return;
		}
```

fails on the first conditional check (the second one is fine).

jdevalk:
## Summary

* Replace Woo Breadcrumbs with Yoast SEO breadcrumbs now works with Storefront theme too.

## Test instructions

* Use Woo + Storefront, enable breadcrumbs, enable replacing breadcrumbs, see the breadcrumbs get replaced.

Fixes #502

### DIFF
--- a/classes/woocommerce-seo.php
+++ b/classes/woocommerce-seo.php
@@ -92,7 +92,7 @@ class Yoast_WooCommerce_SEO {
 			add_filter( 'wpseo_sitemap_urlimages', [ $this, 'add_product_images_to_xml_sitemap' ], 10, 2 );
 
 			// Fix breadcrumbs.
-			$this->handle_breadcrumbs_replacements();
+			add_action( 'send_headers', [ $this, 'handle_breadcrumbs_replacements' ] );
 		}
 
 		add_action( 'admin_enqueue_scripts', [ $this, 'enqueue_scripts' ] );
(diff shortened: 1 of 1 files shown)


dariaknl:
testing done 👍

PR #516, opened 2020-02-06
Fix missing features in product editor

Xyfi:
## Summary
I know the fix seems silly, but the problem is that instead of throwing an error the code execution just stopped at the point where the function was called. This made it hard to pinpoint the exact point of failure.

* Fixes a bug where the internal linking and additional keyphrase functionality are missing from the product edit page.

*

## Test instructions

* Make sure the Additional keyphrases and internal linking are visible and working correctly.

Fixes Yoast/Bugreports#881

### DIFF
--- a/js/src/yoastseo-woo-handle-excerpt-editors.js
+++ b/js/src/yoastseo-woo-handle-excerpt-editors.js
@@ -41,7 +41,7 @@ function getTextEditorContent( elementID ) {
  *
  * @returns {boolean} Whether TinyMCE is available.
  */
-function isTinyMCEAvailable( editorID ) {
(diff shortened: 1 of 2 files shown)


iamazik:
Can you please add a milestone for this?

jdevalk:
@iamazik actually not _this_ pull but the other one that's going to be in 12.6, hence the fact that I didn't put a milestone on it immediately :)

suascat:
Also, could be related to this issue, Facebook and Twitter under the Social tab of Yoast SEO Metabox are not functional when WooCommerce SEO v12.5.1 is enabled and vice versa.

jdevalk:
@suascat can confirm that that's fixed with this merge too.

PR #522, opened 2020-02-06
QA: use the most specific test assertion

jrfnl:
## Summary
* [Unit tests only] QA: use the most specific test assertion

### QA: use the most specific PHPUnit assertion possible

This is a long established best practice.

PHPUnit contains a variety of assertions and the ones available has been extended hugely over the years.

To have the most reliable tests, the most specific assertion should be used.

This implements this for the Clicky plugin, while keeping in mind that at this time, PHPUnit 4.x should still be supported.

Refs:
* https://phpunit.de/manual/4.8/en/appendixes.assertions.html
* https://phpunit.readthedocs.io/en/7.5/assertions.html#

Most notably, this changes calls to `assertEquals()` to `assertSame()`, where `assertEquals()` does a loose type comparison and `assertSame()` does a strict type comparison.

Refs:
* https://phpunit.de/manual/6.5/en/appendixes.assertions.html#appendixes.assertions.assertEquals
* https://phpunit.de/manual/6.5/en/appendixes.assertions.html#appendixes.assertions.assertSame

### Tests: account for test bugs now exposed

Using `assertSame()` exposed a number of bugs in the tests.

The bugs can be classified as:
1. Array order - this is a barely relevant difference in the test cases affected, but does impact any code iterating over an array using `foreach()` instead of directly accessing the array indexes.
2. Type differences in the output - `string` vs `int` etc.

**I've fixed this based on the premise that these are bugs in the _tests_, not bugs in the code.**

**If these are bugs in the _code_, this commit should be partially reverted and the code fixes should be added instead.** /cc @jdevalk 

## Test instructions

* _N/A_
    This is a test-only change and should have no effect on the functionality. If the unit tests still pass, we're good.

### DIFF
--- a/tests/classes/opengraph-test.php
+++ b/tests/classes/opengraph-test.php
@@ -48,14 +48,14 @@ class OpenGraph_Test extends TestCase {
 		);
 
 		$og = new WPSEO_WooCommerce_OpenGraph();
-		$this->assertEquals( 'product', $og->return_type_product( 'article' ) );
+		$this->assertSame( 'product', $og->return_type_product( 'article' ) );
 
 		Functions\stubs(
 			[
 				'is_singular' => false,
 			]
 		);
-		$this->assertEquals( 'article', $og->return_type_product( 'article' ) );
+		$this->assertSame( 'article', $og->return_type_product( 'article' ) );
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


jdevalk:
Looks great, thanks!

PR #525, opened 2020-02-07
Handle permalink changes

### LINKED ISSUE #501: Handle the permalink saves of WooCommerce fields.
We should listen to the option change of `woocommerce_permalinks` and then reset the permalinks for the things that got changed. This need some calculation of difference between the old and new option value. 

Also to integrate fully with https://github.com/Yoast/wordpress-seo/pull/14275 we have to remove the product from the post types array:
```php
\add_filters( 'wpseo_post_types', function( $post_types, $context ) {
   if ( $context === 'update_permalink' ) {
       unset( $post_types['product'] ); 
   }

   return $post_types;
 }, 10, 2 );
```

~We should find a way to make `reset_permalink_indexables` reusable or maybe~ be straight forward and copy the method to the class we make for this pull request.

## To do
- [x] Listen to the option change of `woocommerce_permalinks`
- [x] Check if option has changed: 
   - [x] Reset the permalinks if there is an option change (depending on what value has changed).
- [x] Remove the product from the post types array
- [ ] Write tests

andizer:
## Summary

* N/A - Clears the WooCommerce related indexables

*

## Test instructions

* Change the WooCommerce permalink settings and verify in the database that the permalinks for the related indexables (products, attribute terms, product categories and product tags) are set to `NULL`

Fixes #501


classes/woocommerce-permalink-watcher.php
```diff
@@ -0,0 +1,112 @@
...
+			'tag_base'      => 'product_tag',
+		];
+
+		foreach ( array_keys( $changed_options ) as $changed_term ) {
+			if ( ! isset( $base_to_term_map[ $changed_term ] ) ) {
+				continue;
+			}
+
+			$this->reset_permalink_indexables( 'term', $base_to_term_map[ $changed_term ] );
+		}
```
hansjovis: If the chance of new bases being added to WooCommerce is low, I would simplify this and align it with the other changed option checks:
```php
if ( array_key_exists( 'category_base', $changed_options ) ) {
   $this->reset_permalink_indexables( 'term', 'product_cat' );
}
if ( array_key_exists( 'tag_base', $changed_options ) ) {
   $this->reset_permalink_indexables( 'term', 'product_tag' );
}
```
In my opinion this is less complex and thus a bit easier to comprehend what is happening. (E.g. 'check if `category_base` exists in the changed options, if so, reset the indexables of the `product_cat` sub type').

hansjovis:
Acceptance test: ✅

PR #439, opened 2019-11-15
Prepare for indexables

andizer:
## Summary

This PR can be summarized in the following changelog entry:

* N/A - Makes sure the canonical is fetch the right way.

## Relevant technical choices:

*

## Test instructions

This PR can be tested by following these steps:

* Please test this against the `features/indexables-frontend` branch in Yoast SEO free
* Have WooCommerce active and add a product
* View the product and verify the canonical is in the WooCommerce schema output.

This is part of https://github.com/Yoast/wordpress-seo/issues/13832

PR #528, opened 2020-02-14
Use surface for getting value

andizer:
## Summary

* N/A - Change the way we get the canonical by using the surface.

*

## Test instructions

* Please test this against the features/indexables-frontend branch in Yoast SEO free
* Have WooCommerce active and add a product
* View the product and verify the canonical is in the WooCommerce schema output.
   * In other words: there should be a Schema node with the type `ItemPage`, with an `@id` that is the same as the `<link rel="canonical>` on the page. 

Related to https://github.com/Yoast/wpseo-woocommerce/pull/439

### DIFF
--- a/classes/woocommerce-opengraph.php
+++ b/classes/woocommerce-opengraph.php
@@ -5,6 +5,8 @@
  * @package WPSEO/WooCommerce
  */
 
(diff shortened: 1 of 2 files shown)


hansjovis:
Acceptance test: ✅

Issue #469, opened 2020-01-08
[WIP] Aggregate Offers approach should output individual ProductModel pieces


jono-alderson:
When a product contains multiple variations (e.g., size, color) with varying prices, WooCommerce outputs an AggregateOffer piece, instead of an Offer piece.

This is correct behaviour, but we should also output a `ProductModel` piece (in a `model` array on the `product`) for each variation.

jono-alderson:
Dependent upon https://github.com/Yoast/wpseo-woocommerce/issues/508

Pcosta88:
Please inform the customer of conversation # 610099 when this conversation has been closed.

Pcosta88:
Please inform the customer of conversation # 618056 when this conversation has been closed.

Pcosta88:
Please inform the customer of conversation # 617352 when this conversation has been closed.

priscillamc:
Please inform the customer of conversation # 623115  when this conversation has been closed.

priscillamc:
Please inform the customer of conversation # 642772  when this conversation has been closed.

Pcosta88:
related https://github.com/Yoast/featurerequests/issues/318

Pcosta88:
related https://github.com/Yoast/featurerequests/issues/685

Pcosta88:
Please inform the customer of conversation # 669216 when this conversation has been closed.

Pcosta88:
Please inform the customer of conversation # 698381 when this conversation has been closed.

jono-alderson:
Moved to Jira: https://yoast.atlassian.net/browse/PRODUCT-254

iamazik:
Please inform the customer of conversation # 725430 when this conversation has been closed.

emc2online:
any updates about when this would be available?

monbauza:
Please inform the customer of conversation # 821763 when this conversation has been closed.

PR #532, opened 2020-02-25
Add color property to the product schema output.

afercia:
## Summary

* Adds a `color` attribute to the `Product` Schema output.

Product attributes in WooCommerce are custom attributes users may want to set. As far as I can tell, only `weight` and `dimensions` are default attributes and are part of the product _shipping_. Thus, the only way I could think of to get a product color is by doing the same thing we're doing for `brand` and `manufacturer`.
- adds a `Color` schema addition in SEO > WooCommerce SEO > Schema & OpenGraph additions
- users can create a product attribute and set it to be used as the Color addition
- the color Schema property is output as an array of colors, with lowercase color names
- adds a test for the product color output
- adjust existing tests

## Test instructions

- go to Products > Attributes
- create a new attribute: the attribute name is not relevant, could be "Color" or even "Papede papede"
- click "Configure terms" for the attribute you just created
- create a few attribute terms: these are the actual colors (the color names are not relevant)
- for testing purposes, create also a color with an uppercase name
- go to SEO > WooCommerce SEO
- under "Schema & OpenGraph additions" set the attribute you created as value for the "Color" addition, e.g.:

Screenshot 2020-02-25 at 16 17 02

**Simple product:**

- create a new product of type "Simple product"
- set at least its price and make sure it's in stock so that it's visible in the front end when published
- in the "Product data" meta box under "Attributes" add "Color" as custom product attribute
- click "Select all" under the "Value" textarea to add all the colors
- check the "Visible on the product page" checkbox
- inspect the schema output in the front end
- see the product schema has a `color` property with the colors within an array e.g.:
```
"color": [
    "green",
    "red",
    "uppercasecolor",
    "white"
],
```
- see the color you created with an uppercase name gets output lowercase
- edit the product attributes and remove all colors but one and save
- see the schema output has now just one color
- edit the product, remove all colors and save: Woo removes the Color attribute as well
- see the schema output doesn't have a color property

**Variable product:**

- create a new product of type "Variable product"
- the exact steps are a bit out of the scope of this PR but basically you need to set the color attribute to be used as "product variation" and then set each color as variation
- inspect the schema output in the front end
- see there's a `color` property for the `Product`
- the `Offer` product type will change to `AggregateOffer`
- NOTE: hopefully this is OK otherwise setting the specific color for each `Offer` under `AggregateOffer` would be a bit more complicated /Cc @jono-alderson 

**Grouped product:**

- create a new product of type "Grouped product"
- the exact steps are a bit out of the scope of this PR but basically you need to link the product to another existing product
- inspect the schema output in the front end
- see there's a `color` property for the `Product`

**External/Affiliate product:**
- create a new product of type "External/Affiliate product"
- the exact steps are a bit out of the scope of this PR but basically you need to insert a URL to an external product site 
- inspect the schema output in the front end
- see there's a `color` property for the `Product`

Fixes https://github.com/Yoast/featurerequests/issues/418

### DIFF
--- a/classes/option-woo.php
+++ b/classes/option-woo.php
@@ -61,6 +61,7 @@ if ( ! class_exists( 'WPSEO_Option_Woo' ) && class_exists( 'WPSEO_Option' ) ) {
 			// Form fields.
 			'woo_schema_brand'              => '',
 			'woo_schema_manufacturer'       => '',
+			'woo_schema_color'              => '',
 			'woo_breadcrumbs'               => true,
 			'woo_hide_columns'              => true,
 			'woo_metabox_top'               => true,
@@ -123,6 +124,7 @@ if ( ! class_exists( 'WPSEO_Option_Woo' ) && class_exists( 'WPSEO_Option' ) ) {
 
 					case 'woo_schema_brand':
 					case 'woo_schema_manufacturer':
+					case 'woo_schema_color':
 						if ( isset( $dirty[ $key ] ) ) {
 							if ( in_array( $dirty[ $key ], $valid_taxonomies, true ) ) {
 								$clean[ $key ] = $dirty[ $key ];
--- a/classes/woocommerce-schema.php
+++ b/classes/woocommerce-schema.php
@@ -128,6 +128,7 @@ class WPSEO_WooCommerce_Schema {
 		$this->add_image( $canonical );
 		$this->add_brand( $product );
 		$this->add_manufacturer( $product );
+		$this->add_color( $product );
 		$this->add_global_identifier( $product );
 
 		return [];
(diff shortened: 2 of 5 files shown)


jono-alderson:
Ah, I had a WIP ticket (https://github.com/Yoast/wpseo-woocommerce/issues/469) which suggested we just sniff property names for `colou?r`. That feels more intuitive than forcing the user to map an additional property?

afercia:
The problem with a "sniffing" approach is that users may give a custom attribute a totally unrelated name e.g. `myattribute` or any other random name.

karlijnbok:
CR & Acceptance: 👍

PR #567, opened 2020-04-02
Escape Attribute term

JesserH:
## Context

* We want to users to not be able of entering possible harmful javascript code into the schema using the attributes

## Summary

* Fixes a bug where possible harmful javascript could be entered in the database and affect a website.

* Chosen for `WPSEO_Utils::format_json_encode` over `json_encode` because it is conform the Yoast standards
* Chosen to remove the first and last character of the string if they happen to be double quotes, because otherwise the schema output would include 4 double quotes in total (""example"")

## Test instructions

1. Install Yoast SEO (Premium), Woocommerce and wpseo woocommerce
2. Create a product attribute names `brand` and add a term called `banana`
3. In SEO -> WooCommerce SEO -> set Brand to `Product brand`
4. Create a product, give it a price and assign the brand attribute, set it to use `banana`. Save the product.
5. In your database, go to `<wp-prefix>_terms` and lookup `banana`. Change the name to `</script><script>alert(1)</script><script>`
6. Visit the product on the frontend, note that no pop-up shows
7. Inspect the source code and note that in the product graph the `name` of `"brand"` is ""
8. In your database, change the name to `</script><h1>hello</h1><script>`
9. Visit the product on the frontend, note that no `h1` shows and the additional information is the same as the name in the database 
10. Inspect the source code and note that in the product graph the `name` of `"brand"` is "hello"
11. In your database change the name to `Gimli`
12. Visit the product on the frontend, note that the additional information is the same as the name in the database (Gimli)
13. Inspect the source code and note that in the product graph `name` is "Gimli"
14. In your database change the name to `"Gimli"` (with the double quotes)
15. Visit the product on the frontend, note that the additional information is the same as the name in the database ("Gimli") with quotes
16. Inspect the source code and note that in the product graph the quotes are escaped

Fixes Yoast/bugreports#927

### DIFF
--- a/classes/woocommerce-schema.php
+++ b/classes/woocommerce-schema.php
@@ -300,15 +300,8 @@ class WPSEO_WooCommerce_Schema {
 	 * @param string     $taxonomy  The taxonomy to get the attribute's value from.
 	 */
 	private function add_organization_for_attribute( $attribute, $product, $taxonomy ) {
-		$term = $this->get_primary_term_or_first_term( $taxonomy, $product->get_id() );
-
-		// Escape the string, so harmful input data will not be executed.
-		$escaped_term_name = WPSEO_Utils::format_json_encode( $term->name );
-		// Remove possible double quotes.
-		$stripped_term_name = wp_strip_all_tags( $escaped_term_name );
-		if ( $stripped_term_name[0] === '"' && $stripped_term_name[ ( strlen( $stripped_term_name ) - 1 ) ] === '"' ) {
-			$stripped_term_name = substr( $stripped_term_name, 1, -1 );
-		}
+		$term               = $this->get_primary_term_or_first_term( $taxonomy, $product->get_id() );
+		$stripped_term_name = wp_strip_all_tags( $term->name );
 
 		if ( $term !== null ) {
 			$this->data[ $attribute ] = [
--- a/classes/woocommerce-schema.php
+++ b/classes/woocommerce-schema.php
@@ -274,12 +274,11 @@ class WPSEO_WooCommerce_Schema {
 		if ( $tag_stripped_term_name[0] === '"' && $tag_stripped_term_name[ ( strlen( $tag_stripped_term_name ) - 1 ) ] === '"' ) {
 			$tag_stripped_term_name = substr( $tag_stripped_term_name, 1, -1 );
 		}
-		$term->name = $tag_stripped_term_name;
 
 		if ( $term !== null ) {
 			$this->data[ $attribute ] = [
 				'@type' => 'Organization',
-				'name'  => $term->name,
+				'name'  => $tag_stripped_term_name,
 			];
 		}
 	}
--- a/classes/woocommerce-schema.php
+++ b/classes/woocommerce-schema.php
@@ -300,20 +300,20 @@ class WPSEO_WooCommerce_Schema {
 	 * @param string     $taxonomy  The taxonomy to get the attribute's value from.
 	 */
 	private function add_organization_for_attribute( $attribute, $product, $taxonomy ) {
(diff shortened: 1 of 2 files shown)


classes/woocommerce-schema.php
```diff
@@ -300,12 +300,20 @@ private function add_manufacturer( $product ) {
...
+		$tag_stripped_term_name = wp_strip_all_tags( $escaped_term_name );
+		if ( $tag_stripped_term_name[0] === '"' && $tag_stripped_term_name[ ( strlen( $tag_stripped_term_name ) - 1 ) ] === '"' ) {
+			$tag_stripped_term_name = substr( $tag_stripped_term_name, 1, -1 );
+		}
 
 		if ( $term !== null ) {
 			$this->data[ $attribute ] = [
 				'@type' => 'Organization',
-				'name'  => $term->name,
+				'name'  => $tag_stripped_term_name,
```
karlijnbok, line 316: Maybe rename `$escaped_term_name` and `$tag_stripped_term_name` to `$sanitized_term_name`?
> $sanitized_term_name      = WPSEO_Utils::format_json_encode( $term->name );
		$sanitized_term_name = wp_strip_all_tags( $sanitized_term_name );
		if ( $sanitized_term_name[0] === '"' && $sanitized_term_name[ ( strlen( $sanitized_term_name) - 1 ) ] === '"' ) {
			    $sanitized_term_name = substr( $sanitized_term_name, 1, -1 );
		}

		if ( $term !== null ) {
			$this->data[ $attribute ] = [
				'@type' => 'Organization',
				'name'  => $tag_stripped_term_name,
			];
		}

classes/woocommerce-schema.php
```diff
@@ -300,12 +300,20 @@ private function add_manufacturer( $product ) {
...
 	 */
 	private function add_organization_for_attribute( $attribute, $product, $taxonomy ) {
-		$term = $this->get_primary_term_or_first_term( $taxonomy, $product->get_id() );
+		$term     = $this->get_primary_term_or_first_term( $taxonomy, $product->get_id() );
+		$termname = $term->name;
+
+		// Escape the string so that harmful input data will not be executed.
+		$escaped_term_name      = WPSEO_Utils::format_json_encode( $termname );
+		$tag_stripped_term_name = wp_strip_all_tags( $escaped_term_name );
+		if ( $tag_stripped_term_name[0] === '"' && $tag_stripped_term_name[ ( strlen( $tag_stripped_term_name ) - 1 ) ] === '"' ) {
```
karlijnbok, line 309: Possibly add comment stating something like: "// Remove possible double quotes."

karlijnbok:
Discussed CR with @JesserH, already applied changes.

karlijnbok:
Double quotes are encoded two times with `WPSEO_Utils::format_json_encode`: 
- once in `add_organization_for_attribute` (in `woocommerce-schema.php`), and
- once in `schema_tag`(in `class-wpseo-utils.php`)

Changing the name in the Database to: `"Gimli"` changes the brand name in the schema output to: 
Screenshot 2020-04-03 at 13 59 40

Echoing `$term->name`, `$escaped_term_name` and `$stripped_term_name` results in:
Screenshot 2020-04-03 at 14 14 40

Might be a good idea to look in Free to see how the schema output is ensured and escaped there.

JesserH:
@karlijnbok Updated the logic so the json encode is not triggered in the function itself, which makes every input with a tag (for instance: "</script><script>alert(1)</script><script>") appear on the product page itself but does not render anything in the schema. I think this solves the problem, what are your thoughts?
The entering of "Gimli" (with quotes) in the database conforming the test instructions now renders "\"Gimli\"", which i believe is wished behaviour

karlijnbok:
CR & Acceptance: 👍

Issue #555, opened 2020-03-16
Alterations to product availability meta/schema


jono-alderson:
We need to make some improvements to the logic which determines which meta tags and schema values are output on Woo product pages, specifically relating to product _availability_. 

See https://developer.yoast.com/product-availability-woocommerce/ for further context.

### When a product is in stock
- Output a `og:availability` meta tag.

  E.g.. `<meta property="og:availability" content="instock" />`

### When a product is out of stock
- Add `product:availability` and `og:availability` meta tags, with a value of `out of stock`.

  E.g.. `<meta property="product:availability" content="out of stock" />` and `<meta property="og:availability" content="out of stock" />`

### When a product is out of stock but is set to allow back ordering
- Alter the value of `product:availability` from `in stock` to `available for order`.

  E.g., `<meta property="product:availability" content="available for order" />`

-  Alter the `availability` property of the schema `Offer` from `http://schema.org/InStock` to `https://schema.org/PreOrder`.

-  Add a `og:availability` meta tag with a value of backorder.
  E.g., `<meta property="og:availability" content="backorder" />`

afercia:
To clarify: 

> When a product is in stock
> Output a og:availability meta tag.

I read this as we should output _only_ the `og:availability` meta tag.

> When a product is out of stock but is set to allow back ordering
> Alter the value of product:availability from in stock to available for order.

Instead, this sounds like a `product:availability` > `in stock` is there.

I'm assuming we should output both, correct?

jono-alderson:
Yep. These requirements should be read as 'in addition to current behaviour', except for where instructions alter existing behaviour.

igorschoester:
Please note the output of all this got changed to presenters for the indexables in the upcoming 13.0 release in #568

PR #568, opened 2020-04-02
Bring back product metatags

### LINKED ISSUE #566: Adds logic to bring back the og metatags #802
With the upcoming release of Yoast SEO 14.0 the filter wpseo_opengraph is deprecated. Woocommerce SEO is hooking into this filter for rendering OG metatags. We should replaces these by presenters.

For reference see:
https://github.com/Yoast/wpseo-woocommerce/blob/c0ea263345da47cb3b58c3855a5f03ee3cf5b0da/classes/woocommerce-opengraph.php#L21

https://github.com/Yoast/wpseo-woocommerce/blob/c0ea263345da47cb3b58c3855a5f03ee3cf5b0da/classes/woocommerce-opengraph.php#L72

igorschoester:
## TODO

* Write tests

## Context

* We should...
  1. ...still output our metatags as before
  2. ...not trigger our own deprecation notices
  3. ...show deprecation notices when the functions are still called by external parties

## Summary

* Adds a WooCommerce presenter to bring back the product metatags that relied on the (deprecated) `wpseo_opengraph` action.

* I'm not sure about...
  * ...the location of the deprecation of the `Yoast\WP\Woocommerce\opengraph` action. I put it there because the old location (`product_enhancement`) is deprecated too.
  * ...the removal of the action in the old `product_enhancement` method.
* I chose not to update too much, if you spot some easy fixes, feel free to apply them 😉 

## Test instructions

* Install WooCommerce and WooCommerce Brands and set up a product with a brand, price, availability and so on.
* Compare the metatag output before this branch and after this branch, there should be:
  - product:brand
  - product:price:amount
  - product:price:currency
  - og:availability
  - product:availability
  - product:retailer_item_id
  - product:condition
* The metatag of `product:availability` when in stock is changed compared to what it was before. From `in stock` to `instock`. As this got noted as being wrong by Jono, see: https://github.com/Yoast/wpseo-woocommerce/issues/555 for future changes.
* You could manually run the actions that are now deprecated to see the notices.

Fixes #566 
Fixes https://github.com/Yoast/bugreports/issues/950


classes/woocommerce-presenter.php
```diff
@@ -0,0 +1,195 @@
...
+		}
+
+		/**
+		 * Action: 'Yoast\WP\Woocommerce\opengraph' - Allow developers to add to our OpenGraph tags.
+		 *
+		 * @since 12.6.0
+		 *
+		 * @api   WC_Product $product The WooCommerce product we're outputting for.
+		 */
+		do_action_deprecated( 'Yoast\WP\Woocommerce\opengraph', $product, 'WPSEO Woo 13.0', 'WPSEO_WooCommerce_Presenter' );
```
andizer, line 33: Why is this deprecated? This means there is no way for third parties to output extra OG-tags?

herregroen: I discussed this with Igor and felt that the existing filters should be sufficient.

The only thing that's wrong here is the replacement. That should be the `wpseo_frontend_presenters` filter.

hansjovis:
Acceptance test: 🚫 

When I try to view the product on the frontend, I get a white screen, nothing else 😕. 
(With `release/14.0` of `wordpress-seo` checked out).

When I deactivate all other plugins except WooCommerce, Wordpress SEO and WooCommerce SEO I get this fatal error:

```
Fatal error: Class WPSEO_WooCommerce_Presenter contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (Yoast\WP\SEO\Presenters\Abstract_Indexable_Presenter::get)
```

The `Abstract_Indexable_Presenter` has changed last Thursday. It now expects a `get` method to be implemented. I expect Herre has more context on what has changed.

classes/woocommerce-abstract-product-availability-presenter.php
```diff
@@ -0,0 +1,40 @@
...
+
+	/**
+	 * Whether the product is in stock.
+	 *
+	 * @var bool
+	 */
+	protected $is_in_stock;
+
+	/**
+	 * WPSEO_WooCommerce_Abstract_Product_Presenter constructor.
```
andizer, line 28:
```suggestion
	 * WPSEO_WooCommerce_Abstract_Product_Availability_Presenter constructor.
```

classes/woocommerce-opengraph.php
```diff
@@ -154,6 +139,8 @@ protected function get_brand_term_name( $schema_brand, $product ) {
 	 * @param WC_Product $product The WooCommerce product object.
 	 */
 	public function brand( WC_Product $product ) {
+		_deprecated_function( __METHOD__, 'WPSEO Woo 13.0', 'WPSEO_WooCommerce_Presenter::product_brand' );
```
andizer, line 142: Don't forget to replace this and the other ones referring to `WPSEO_WooCommerce_Presenter` with the actual presenters.

classes/woocommerce-product-availability-presenter.php
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * WooCommerce Yoast SEO plugin file.
+ *
+ * @package WPSEO/WooCommerce
+ */
+
+use WPSEO_WooCommerce_Abstract_Product_Availability_Presenter;
+
+/**
+ * Represents the product's price currency.
```
johannadevos, line 11:
```suggestion
 * Represents the product's availability.
```

classes/woocommerce-product-retailer-item-id-presenter.php
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * WooCommerce Yoast SEO plugin file.
+ *
+ * @package WPSEO/WooCommerce
+ */
+
+use WPSEO_WooCommerce_Abstract_Product_Presenter;
+
+/**
+ * Represents the product's retail item ID.
```
johannadevos, line 11:
```suggestion
 * Represents the product's retailer item ID.
```

classes/woocommerce-product-availability-presenter.php
```diff
@@ -0,0 +1,38 @@
...
+	protected $tag_format = '<meta property="product:availability" content="%s" />';
+
+	/**
+	 * Gets the raw value of a presentation.
+	 *
+	 * @return string The raw value.
+	 */
+	public function get() {
+		if ( $this->is_on_backorder ) {
+			return 'available for order';
```
johannadevos, line 29: In `classes/woocommerce-pinterest-product-availability-presenter.php` you return `'backorder'`, here you return `'available for order'`. For consistency, should it not be the same string?

classes/woocommerce-seo.php
```diff
@@ -1220,4 +1239,40 @@ public function upgrade() {
...
+	 *
+	 * @return bool True when the price should be shown.
+	 */
+	private function should_show_price() {
+		/**
+		 * Filter: wpseo_woocommerce_og_price - Allow developers to prevent the output of the price in the OpenGraph tags.
+		 *
+		 * @deprecated 12.5.0. Use the {@see 'Yoast\WP\Woocommerce\og_price'} filter instead.
+		 *
+		 * @api        bool unsigned Defaults to true.
```
johannadevos, line 1254:
```suggestion
		 * @api bool unsigned Defaults to true.
```

classes/woocommerce-seo.php
```diff
@@ -1220,4 +1239,40 @@ public function upgrade() {
...
+			'Yoast WooCommerce 12.5.0',
+			'Yoast\WP\Woocommerce\og_price'
+		);
+
+		/**
+		 * Filter: Yoast\WP\Woocommerce\og_price - Allow developers to prevent the output of the price in the OpenGraph tags.
+		 *
+		 * @since 12.5.0
+		 *
+		 * @api   bool unsigned Defaults to true.
```
johannadevos, line 1268:
```suggestion
		 * @api bool unsigned Defaults to true.
```

classes/woocommerce-product-availability-presenter.php
igorschoester: No, unfortunately, this is not in our control. It is how they implemented their metatags.

johannadevos:
CR: ✅ , except there is one unfinished task. 
Note: I only reviewed commits `240aff2` and `a77480b` because @hansjovis already did a CR of the rest.
* [x] The `meta` surface is now called correctly.
* [x] The refactoring of the presenters looks good to me.
* [x] All the new presenters are added in `woocommerce-seo.php`. 
* [x] I pushed three commits with small fixes and improvements.
* [ ] @igorschoester will finish replacing the presenters in the deprecated functions.

hansjovis:
Acceptance test: 🚫 

The value of the `product:availability` meta tag is different when the product is in stock.
In this branch it is `instock`, whereas it is `in stock` in `release/13.0`:
```html
<meta property="product:availability" content="instock">
```
vs
```html
<meta property="product:availability" content="in stock">
```

It works for the rest 🙂.

hansjovis:
> The value of the product:availability meta tag is different when the product is in stock.
In this branch it is `instock`, whereas it is `in stock` in release/13.0

This works as expected, as noted by Jono (see #555 for future changes).

PR #576, opened 2020-04-05
QA: use the variable

jrfnl:
## Context
* Minor performance improvement

## Summary
* Minor performance improvement

Above the `foreach` the `$prices_include_tax` variable is defined with the output of the same function call, so we may as well use it and remove the redundant repeated function call which would always have the same output anyway.

:arrow_right:  _Leaving this open for someone to quickly look this over in case I'm missing some context and the return value of the function call could change during the running of the loop._

## Test instructions

* This PR should have no impact on the functionality, so the schema code added for offers should be the same before and after.


bintzandt:
The variable stays the same unless someone manages to change the option while the schema is being loaded.

That seems like a very small chance, and would be fixed with a reload. Therefore, I don't see any problems here either.

Merging.

PR #591, opened 2020-04-15
Fix OpenGraph deprecations

igorschoester:
## Context

* Don't deprecate the whole class as there are still things in use.

## Summary

*  Fixes the OpenGraph class deprecations.

* Removed the tests for the deprecated methods.

## Test instructions

* Check if all the methods with `_deprecated_function` have the doc comments `@deprecated`.
* Check that the deprecated methods are not used anywhere anymore (except the protected helper one: `get_brand_term_name`).

Fixes #


classes/woocommerce-opengraph.php
```diff
@@ -115,12 +115,17 @@ public function product_namespace( $input ) {
...
+	 * @deprecated 13.0
+	 * @codeCoverageIgnore
+	 *
 	 * @param string      $schema_brand The taxonomy the site uses for brands.
 	 * @param \WC_Product $product      The product we're finding the brand for.
 	 *
 	 * @return bool|string The brand name or false on failure.
 	 */
 	protected function get_brand_term_name( $schema_brand, $product ) {
+		_deprecated_function( __METHOD__, 'WPSEO Woo 13.0', 'WPSEO_WooCommerce_Product_Brand_Presenter::get' );
```
andizer, line 127: Can you make sure every other deprecation notice will have `::get`. I've found some of them, but I can place a comment because the lines where `::get` is missing aren't touched by this pull.

johannadevos:
CR & Acceptance: 👍

PR #595, opened 2020-04-23
Omit the price amount and currency meta for variable and grouped products.

afercia:
## Context

* Avoid to output the lowest price and currency in the og meta `product:price:amount` for variable and grouped products.

## Summary

* Omits the `product:price:amount` and `product:price:currency` og meta for variable and grouped products.

* introduced a utility method `get_product_type()`
* used that in the price amount presenter

## Test instructions

- create a variable product with variations with different prices
- on trunk:
- see the product on the front end
- inspect the source and observe the meta `product:price:amount` outputs the lowest price
- observe there's a `product:price:currency` meta
- switch to this branch
- inspect the source and check the meta `product:price:amount` is omitted
- check the `product:price:currency` is omitted as well
- check the above also for a _grouped_ product 

## Questions
- ~This PR doesn't touch the Schema price piece, which seems correct to me. Is that OK?~
(This has an answer on the issue so can be ignored, see https://github.com/Yoast/featurerequests/issues/112#issuecomment-618409025)
- Should the currency meta be omitted as well? [Edit: addressed]
- What about _grouped_ products? They actually have a price _range_ but the meta outputs only the lowest price [Edit: addressed]

Cc @jono-alderson @jdevalk 

Fixes https://github.com/Yoast/featurerequests/issues/112

### DIFF
--- a/classes/presenters/woocommerce-product-price-amount-presenter.php
+++ b/classes/presenters/woocommerce-product-price-amount-presenter.php
@@ -23,6 +23,13 @@ class WPSEO_WooCommerce_Product_Price_Amount_Presenter extends WPSEO_WooCommerce
 	 * @return string The raw value.
 	 */
 	public function get() {
+		$product_type = WPSEO_WooCommerce_Utils::get_product_type( $this->product );
+
+		// Omit the price amount for variable products.
+		if ( $product_type === 'variable' ) {
+			return '';
+		}
+
 		return (string) WPSEO_WooCommerce_Utils::get_product_display_price( $this->product );
 	}
 }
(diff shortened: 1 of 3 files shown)


afercia:
Additional feedback to address:

As there's no support for ranged/varying costs in meta:
- omit the currency stuff, too
- same for grouped stuff, too

karlijnbok:
CR & Acceptance: 👍

FITZKRAWALLDO:
Can somebody explain to me, why you removed this microtags without replacing it? So for me, not having microtags make yoast completely unusable for e.g. facebook marketing api. For facebook "price" is a must - on all products

PR #596, opened 2020-04-28
Don't show open graph images gallery when image set by user

afercia:
## Context

* We want to improve the open graph image output when users set a specific image for a product.

## Summary

* Fixes a bug where all product gallery images were used as open graph images even though users set a specific image for a product.

* Introduces a new method to check whether users set a specific image for a product, based on the `indexables` implementation. Hat tip to @hansjovis 
* Refactors most of the related tests to use a mocked instance of the main class to test.

## Test instructions

- install Yoast SEO, WooCommerce, Yoast SEO: WooCommerce
- on trunk
- create a product and add some images in the "Product gallery" meta box
- add an image in the Yoast SEO meta box > Social > Facebook > Facebook Image
- publish the product
- see the product on the front end and inspect the source
- see there are multiple images used as open graph images: all the ones from the product gallery plus the one set in the Yoast SEO metabox
- switch to this branch 
- refresh the product page on the front end and inspect the page source
- check that only the image set in the Yoast SEO metabox is now used as open graph image
- remove the image set in the Yoast SEO metabox and update the product post
- inspect the product page source again and see that only the images from the gallery are now used as open graph images

## Questions
- I'd like this behavior be confirmed as the desired one /Cc @jono-alderson @jdevalk 
- the refactored tests should be reviewed, specifically: wondering if there’s a specific reason why the tests were instantiating the `WPSEO_WooCommerce_OpenGraph` class each time. I changed that to use a mocked class instance so that I can use `shouldReceive( 'is_opengraph_image_set_by_user' )`

Fixes https://github.com/Yoast/bugreports/issues/812

### DIFF
--- a/classes/woocommerce-opengraph.php
+++ b/classes/woocommerce-opengraph.php
@@ -318,6 +318,11 @@ class WPSEO_WooCommerce_OpenGraph {
 	 * @return bool True on success, false on failure.
 	 */
 	protected function set_opengraph_image_product( $opengraph_image, WC_Product $product ) {
+		// Don't add the gallery images if the user set a specific image for this product.
+		if ( $this->is_opengraph_image_set_by_user( $product->get_id() ) ) {
+			return true;
+		}
+
 		$img_ids = $product->get_gallery_image_ids();
 
 		if ( is_array( $img_ids ) && $img_ids !== [] ) {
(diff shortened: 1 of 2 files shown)


jono-alderson:
Sounds good to me!

classes/woocommerce-opengraph.php
```diff
@@ -331,6 +336,20 @@ protected function set_opengraph_image_product( $opengraph_image, WC_Product $pr
 		return false;
 	}
 
+	/**
+	 * Checks whether users set a specific open graph image for a product.
```
karlijnbok, line 340: Should `users` be plural? Or should it be `user has` / `user's`, or simply `user`? Just checking :)

classes/woocommerce-opengraph.php
```diff
@@ -331,6 +336,20 @@ protected function set_opengraph_image_product( $opengraph_image, WC_Product $pr
...
+	 *
+	 * @param int $product_id The product ID.
+	 *
+	 * @return bool Whether users set a specific open graph image for a product.
```
karlijnbok, line 344: See comment above

classes/woocommerce-opengraph.php
afercia: Ah well it's just a matter of preference I guess :) Generally, I don't like to use the singular form `the user` because it sounds like an abstract entity, like if the "the user" isn't a human being. Instead, `users` sounds more like a group of persons.

karlijnbok:
CR & Acceptance: 👍

PR #607, opened 2020-05-20
Alter schema availability for products on backorder.

afercia:
## Context

* We want to improve the Product Schema to make the `availability` property accurate.

## Summary

* Outputs a `PreOrder` value for the Product Schema availability property when the Product is "on backorder".

- filtered the availability value in the `filter_offers()` method
- added a test for the "on backorder" case and adjusted the existing tests

## Test instructions

- create or edit a simple product
- in "Product data" meta box > Inventory, set the "Stock status" to "On backorder"
- view the product on the front end
- append the URL parameter `?yoastdebug` to the URL in your browser's address bar and press Enter to request again the page
- see the page source (on macOS Chrome: Alt + Cmd  U)
- see the product schema availability property
- check its value is `"availability": "http://schema.org/PreOrder"`

Note:
WooCommerce has 4 different types of products and only the "simple" product provide a UI in the 
"Product data" meta box to change the stock availability.

For the WooCommerce "variable" and "grouped" products:
- the UI to set stock availability is intentionally hidden, see the CSS classes on the "stock status" UI element `stock_status_field hide_if_variable hide_if_external hide_if_grouped form-field _stock_status_field`
- but then it can be se in the Quick Edit form 🙈 
- so you can change the main product availability in the Quick Edit form
- inspect the products page source on the front end and check the schema availability is `"availability": "http://schema.org/PreOrder"`
- the single offers grouped under the main product don't have an availability property and this is just how WooCommerce works

For the WooCommerce external/affiliate product:
- there's no way to set stock availability, not even in the Quick Edit form, so it's always "in stock"

Fixes https://github.com/Yoast/wpseo-woocommerce/issues/555

### DIFF
--- a/tests/classes/schema-test.php
+++ b/tests/classes/schema-test.php
@@ -231,6 +231,77 @@ class Schema_Test extends TestCase {
 		$product->expects( 'get_price' )->once()->andReturn( 49 );
 		$product->expects( 'get_min_purchase_quantity' )->once()->andReturn( 1 );
 		$product->expects( 'is_on_sale' )->once()->andReturn( false );
+		$product->expects( 'is_on_backorder' )->once()->andReturn( false );
+
+		$this->meta
+			->expects( 'for_current_page' )
+			->once()
+			->andReturn( (object) [ 'site_url' => 'https://example.com/' ] );
+
+		$output = $schema->filter_offers( $input, $product );
+
+		$this->assertSame( $expected_output, $output );
+	}
+
+	/**
+	 * Test filtering offers with produc on backorder.
+	 *
+	 * @covers ::filter_offers
+	 */
+	public function test_filter_offers_with_product_on_backorder() {
+		$schema = new Schema_Double();
+		$input  = [
+			'@type'       => 'Product',
+			'name'        => 'Customizable responsive toolset',
+			'url'         => 'https://example.com/product/customizable-responsive-toolset/',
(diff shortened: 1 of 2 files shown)


afercia:
Node: I did see what Codeclimate complains about. In order to "fix" that the new check should be moved out to a new method _and add one more foreach loop_, which I think should be avoided.

tests/classes/schema-test.php
```diff
@@ -231,6 +231,77 @@ public function test_filter_offers() {
...
+			->once()
+			->andReturn( (object) [ 'site_url' => 'https://example.com/' ] );
+
+		$output = $schema->filter_offers( $input, $product );
+
+		$this->assertSame( $expected_output, $output );
+	}
+
+	/**
+	 * Test filtering offers with produc on backorder.
```
jcomack, line 247: Missing letter `t` for `product`.

tests/classes/schema-test.php
```diff
@@ -231,6 +231,77 @@ public function test_filter_offers() {
...
+						'url'   => 'https://example.com',
+					],
+				],
+			],
+		];
+
+		$expected_output                     = $input;
+		$expected_output['offers'][0]['@id'] = 'https://example.com/#/schema/offer/209643-0';
+
+		$base_url = 'http://example.com';
```
jcomack, line 285: `$base_url` is being defined as a variable, but is only used once. Why not directly set it in the stub on line 288?

afercia: Good spot. Other tests suffer from the same copy-paste issue. Will fix.

jcomack:
Acceptance done 👍

PR #608, opened 2020-05-22
Sku incorrect value to blank

JesserH:
## Context

* When the SKU of a product is left empty, WooCommerce makes it the value of the product's id. 

## Summary

* Fixes a bug where an empty SKU fallbacks to the product's ID.

* Created a `filter_sku` function that is called in `change_product` in `woocommerce-schema.php`

## Test instructions

* Checkout this branch and build
* Install Yoast SEO and Woocommerce
* Add a product (variable or simple)
* Leave SKU blank
* See that no SKU value is printed in the product graph in the source code
* Repeat the above steps on trunk and note that the sku value should be put out in the graph. 

Fixes https://github.com/Yoast/bugreports/issues/357


afercia:
CR & acceptance 👍

PR #578, opened 2020-04-07
Hide all Yoast columns in the Product overview page except the SEO Score column

afercia:
## Context

* Simplify the plugin settings by prodiving sane defaults

## Summary

This PR can be summarized in the following changelog entry:

* Hides all Yoast columns in the Product overview page except the SEO Score column.

## Relevant technical choices:

- removes `woo_hide_columns` from the `wpseo_woo` option in the database
- removes related code used to validate the `woo_hide_columns` option
- adjusts existing test to check all columns except SEO Score aren't displayed

## Test instructions

This PR can be tested by following these steps:
- activate and install Yoast SEO, WooCommerce, and WooCommerce SEO switched to this branch
- check in the database table `wp_options` > option_name `wpseo_woo` there's no `woo_hide_columns` 
- in the admin, go to SEO > WooCommerce SEO and see there's no UI to hide columns any longer (refer to the screenshot on the issue)
- go to the WooCommerce Products overview page and see all the Yoast table columns (focus keyword, title, description, etc.) aren't displayed except the SEO Score one
- click on "Screen Options" at the top right of the page
- uncheck / check the `SEO score` checkbox to hide / unhide the SEO Score column and check it works as expected

## Note
When merging, ask team blog in collaboration with support to write a help article about how to use the Screen options to hide these columns. /cc @tacoverdo @willemientje

Fixes https://github.com/Yoast/featurerequests/issues/464

PR #616, opened 2020-06-15
P3-35 Make the Yoast SEO columns hidden by default

afercia:
## Context

* We want the Yoast SEO columns hidden by default (except SEO score) and users still able to make the visible.

Follow-up to https://github.com/Yoast/wpseo-woocommerce/pull/578/

## Summary

* Makes the Yoast SEO columns hidden by default in the Products overview table.

The main point for this change is that the columns visibility is a _per-user_ setting. In order to make users able to set the columns visibility, this must be saved (and checked against) a user option. I used these WordPress functions:
- get_current_user_id()
- get_user_option()
- update_user_option()

I'd like to discuss with a senior dev / architect if this approach is OK. Also, this makes integration-tests a bit pointless, I guess.

## Test instructions

- activate Yoast SEO, WooCommerce, and Woo SEO
- delete from your database any user meta with meta_key `manageedit-productcolumnshidden` for any user (search for it in the wp_usermeta table)
- go to the Products page
- observe the SEO score column is visible by default and all the other Yoast SEO columns are hidden by default
- expand the Screen Options and verify the related checkboxes state is as the one in the screenshot below:

Screenshot 2020-06-15 at 15 18 42

- check a couple checkboxes to make some of the Yoast SEO columns visible, e.g.: "Readability score" and "Keyphrase"
- navigate to another page
- navigate back to the Products overview
- observe the "Readability score", "Keyphrase", and "SEO score" columns are still visible
- play with the columns, navigate away again, navigate back, check the columns visibility is correct 
- in your database `wp_usermeta` table, delete your user meta with the meta key `wpseo_woo_columns_hidden_default`
- go to the Products page (or refresh the page) and verify the SEO score column is visible and all the other Yoast SEO columns are hidden
- repeat the steps above using another user
- if you need to repeat the test, you will also need to delete from your database `wp_usermeta` table the user meta with meta_key `wpseo_woo_columns_hidden_default`

Fixes [P3-35]

### DIFF
--- a/classes/woocommerce-seo.php
+++ b/classes/woocommerce-seo.php
@@ -536,12 +536,16 @@ class Yoast_WooCommerce_SEO {
 	 * @return void
 	 */
 	public function set_yoast_columns_hidden_by_default( $current_screen ) {
-		$user_id             = get_current_user_id();
-		$user_hidden_columns = get_hidden_columns( $current_screen );
+		$user_id = get_current_user_id();
 
-		if ( get_user_option( 'wpseo_woo_columns_hidden_default', $user_id ) === '1' ) {
+		if (
+			$current_screen->id !== 'edit-product' ||
+			get_user_option( 'wpseo_woo_columns_hidden_default', $user_id ) === '1'
+		) {
 			return;
 		}
+
+		$user_hidden_columns = get_hidden_columns( $current_screen );
 
 		$yoast_hidden_columns = [
 			'wpseo-title',
--- a/classes/woocommerce-seo.php
+++ b/classes/woocommerce-seo.php
@@ -65,8 +65,8 @@ class Yoast_WooCommerce_SEO {
 			add_filter( 'wpseo_submenu_pages', [ $this, 'add_submenu_pages' ] );
 			add_action( 'admin_print_styles', [ $this, 'config_page_styles' ] );
 
-			// Products tab columns.
-			add_filter( 'manage_product_posts_columns', [ $this, 'column_heading' ], 11, 1 );
+			// Hide the Yoast SEO columns in the Product table by default, except the SEO score column.
+			add_action( 'admin_init', [ $this, 'set_yoast_columns_hidden_by_default' ] );
 
 			// Move Woo box above SEO box.
 			add_action( 'admin_footer', [ $this, 'footer_js' ] );
@@ -529,16 +529,18 @@ class Yoast_WooCommerce_SEO {
 	}
 
 	/**
-	 * Removes the Yoast SEO columns in the edit products page.
+	 * Hides the Yoast SEO columns in the Product table by default, except the SEO score one.
 	 *
-	 * @since 1.0
-	 *
-	 * @param array $columns List of registered columns.
-	 *
-	 * @return array Array with the filtered columns.
+	 * @return void
 	 */
-	public function column_heading( $columns ) {
-		$keys_to_remove = [
+	public function set_yoast_columns_hidden_by_default() {
+		$user_id = get_current_user_id();
+
+		if ( get_user_option( 'wpseo_woo_columns_hidden_default', $user_id ) === '1' ) {
+			return;
+		}
+
+		$yoast_hidden = [
 			'wpseo-title',
 			'wpseo-metadesc',
 			'wpseo-focuskw',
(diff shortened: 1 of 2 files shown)


afercia:
To recap the points where I'd like to have feedback from seniors: given that (as far as I see) Woo SEO:
- doesn't use "real" upgrade routines
- doesn't use an activation hook so that it's not easy to run stuff _once_ on activation

Then, I opted to use WordPress functions. Also, the table columns visibility is a WordPress _per-user_ setting saved as a usermeta. Thus, I'm using:
- get_current_user_id() to get the user id
- update_user_option() to update the columns visibility 
- get/update_user_option() to set/check _one more, new_ user meta `wpseo_woo_columns_hidden_default` which is used as a "flag" 

Also, since the new method doesn't return anything and just uses WordPress functions, I'd tend to think integration tests aren't necessary.

classes/woocommerce-seo.php
```diff
@@ -529,32 +529,31 @@ public function footer_js() {
...
-		}
-
-		foreach ( $keys_to_remove as $key_to_remove ) {
-			unset( $columns[ $key_to_remove ] );
+			$yoast_hidden[] = 'wpseo-' . WPSEO_Link_Columns::COLUMN_LINKS;
+			$yoast_hidden[] = 'wpseo-' . WPSEO_Link_Columns::COLUMN_LINKED;
 		}
 
-		return $columns;
+		update_user_option( $user_id, 'manageedit-productcolumnshidden', $yoast_hidden, true );
```
herregroen, line 555: Shouldn't this first be merged with the existing value of this user option?

Also, if the user has no option for this then it doesn't need to be set either.

herregroen: Other than that I think the approach here is good!

afercia: > Also, if the user has no option for this then it doesn't need to be set either.

Seems to me that if there's no `manageedit-productcolumnshidden` user option, then:
- if the SEO and Readability analysis are enabled, their columns will be displayed
- if the "Text link counter" feature is enabled, the two links columns will be displayed

Instead, we want to hide all columns except the SEO score one. Thus, I think this user option needs to be set to actually hide the columns. Unless I'm missing something 🙂

afercia: > Shouldn't this first be merged with the existing value of this user option?

Hm for the Yoast columns, I think no. We want to initially hide all columns except the SEO score one. A potential issue would be if _other_ plugins add their own columns to the Products page? Should we take this scenario into consideration? If so, yes:  _hidden_ columns potentially added by other plugins should be merged with the Yoast hidden columns before saving the user option.

afercia: Hm well I guess also other _default_ hidden columns need to be merged. 🙂

JesserH:
CR 👍 
Acceptance 🏗️ 

At the testing instructions, this step: "observe the SEO score column is visible by default and all the other Yoast SEO columns are hidden by default" is not working. I figured that even if all `manageedit-productcolumnshidden` and `wpseo_woo_columns_hidden_default` fields are deleted, if you have other more or less Yoast SEO columns active then expected, on switch-build-reactivate to the `P3-35-woo-seo-columns-hidden-default` branch they are not correctly set to the expected single Yoast SEO score column. The visible set columns of Yoast SEO are displayed as configured before.

afercia:
A misbehavior was triggered by the fact the code ran on all pages. The last commit should fix it by running only on the Products page. 

Please re-test it **starting from trunk;**
- delete from database the user meta with meta_key `wpseo_woo_columns_hidden_default` if it exists
- go in the Products page and hide a couple Woo columns e.g.:

Screenshot 2020-06-25 at 15 02 03

- navigate to another page
- switch to this branch (I think it's not necessary to build Woo SEO, but in any case... build!) 
- make sure again there's no `wpseo_woo_columns_hidden_default` user meta in your db 
- navigate to some other page e.g. the Posts page and hide some columns there
- then go to the Products page
- observe the Woo columns you previously hid are still hidden 
- observe all the Yoast columns are hidden by default except the SEO score one, e.g.

Screenshot 2020-06-25 at 15 25 45

- navigate to the Posts page and hide some columns there
- go to the Products page and check your columns settings are preserved 
- check your database now has a `wpseo_woo_columns_hidden_default` user meta for your users
- repeat the test with another user

JesserH:
CR and Acceptance ✅

dariaknl:
To recap the whole behaviour:

* **clean install Yoast SEO 14.6 RC1 + Woocommerce (all default settings)**
Screenshot 2020-07-16 at 06 26 41
Screenshot 2020-07-16 at 06 27 35

* **clean install Yoast SEO 14.6 RC1 + Woocommerce  + SEO Woo 13.4 RC1 (all default settings)** 
Screenshot 2020-07-16 at 06 33 09
Screenshot 2020-07-16 at 06 33 16

* ** Yoast SEO 14.5  + Woocommerce + SEO Woo 13.3**
Screenshot 2020-07-16 at 07 00 24

Then I disable SEO score and upgrade SEO Yoast to 14.6 RC1:
Screenshot 2020-07-16 at 07 04 26

Then I install SEO Woo 13.4 RC1
Screenshot 2020-07-16 at 07 06 51

PR #617, opened 2020-06-18
PHPCS: enable caching between runs

jrfnl:
## Context

* Dev environment improvement

## Summary

* Enable PHPCS caching

Since PHPCS 3.0.0, PHPCS has a caching feature available. In this PR, this caching feature is now turned on for this repo.

This means in effect that the very first run of PHPCS with this feature will be marginally slower (on any individual machine). Most runs _after_ that though, will be significantly faster as PHPCS will only scan the files which have changed since the previous run.

The cache will automatically be invalidated and rebuild whenever relevant, like when the version of one of the PHPCS dependencies has changed or when PHPCS itself has been changed.

To (temporarily) turn caching off, run PHPCS like so:
```bash
composer check-cs -- --no-cache
```

The cache file is set to be saved to a `.cache` directory to allow Travis to benefit from this feature as well.
As this directory _has_ to exist for the cache file to be written to it, it is created and committed to the repo with a `.gitkeep` file

### Commit details:

`/.cache/` directory:
* Create the directory and place a `.gitkeep` file in it to allow it to be committed.

PHPCS config file:
* Enable the caching feature to a fixed file called `/.cache/phpcs.cache`.

`.gitignore`:
* Prevent committing of the cache file.

`.gitattributes`:
* Do not include the `.cache` directory in distributable archives.

Travis:
* Use the Travis caching feature to cache the new `.cache` directory (including the cache file) between runs to allow Travis to benefit from the PHPCS caching as well.

## Test instructions

* Check out this branch.
* Run `composer check-cs` and take note of the time the run has taken at the bottom of the report.
* Verify that the `.cache` directory on your machine now contains a `phpcs.cache` file.
* Run `composer check-cs` again. If all went well, the run should be significantly faster compared to the earlier run.

Also verify that Travis correctly caches the `.cache` directory. While that won't make a difference for this build, it should for future builds.

### DIFF
--- a/.cache/.gitkeep
+++ b/.cache/.gitkeep
@@ -1,0 +1,1 @@
+# This directory has to exist to allow cache files to be written to it.
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,6 +5,7 @@
 # https://www.reddit.com/r/PHP/comments/2jzp6k/i_dont_need_your_tests_in_my_production
 # https://blog.madewithlove.be/post/gitattributes/
 #
+.cache export-ignore
 .github export-ignore
 grunt export-ignore
 tests export-ignore
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ artifact.zip
 artifact/
 .phpcs.xml
 phpcs.xml
+.cache/phpcs.cache
 phpunit.xml
 yarn-error.log
 /js/dist/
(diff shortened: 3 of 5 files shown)


bintzandt:
CR: Looks good. Minor changes, not much to comment.
Acc:
- The first run does indeed run marginally slower (818ms vs 855ms)
- Subsequent builds run a lot faster (100ms vs 855ms)
- There is a slight increase in memory usage (which is to be expected): 14MB vs 16MB

Merging :heavy_check_mark:

PR #623, opened 2020-07-02
Update checkbox to new styling

rolf-yoast:
## Context

* Updates the plugin to the new Yoast styling.

## Summary

*  Updates the plugin to the new Yoast styling.

*

## Test instructions

* Build the plugin and get the `https://github.com/Yoast/wpseo-woocommerce/tree/FRO-211-woocommerse-styling-update branch` from the wordpress-seo repo.
* Install and activate Woocommerce
* Go the the Woocommerce page under SEO
* Check if the page uses the new styling

Fixes https://yoast.atlassian.net/browse/FRO-211

### DIFF
--- a/classes/woocommerce-seo.php
+++ b/classes/woocommerce-seo.php
@@ -476,7 +476,7 @@ class Yoast_WooCommerce_SEO {
 			);
 			echo '</p>';
 
-			echo '<fieldset id="id-name" class="yoast-field-group">';
+			echo '<fieldset class="yoast-field-group">';
 			Yoast_Form::get_instance()->checkbox(
 				'woo_breadcrumbs',
 				sprintf(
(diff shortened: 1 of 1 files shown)


classes/woocommerce-seo.php
```diff
@@ -454,6 +454,8 @@ public function admin_panel() {
 			$taxonomies[ strtolower( $object_taxonomy->name ) ] = esc_html( $object_taxonomy->labels->name );
 		}
 
+		echo '<div class="yoast-feature">';
```
JoseeWouters, line 457: This class was meant to be used only on the General - Features page. Maybe we need to make this classname a bit more generic.

classes/woocommerce-seo.php
```diff
@@ -472,8 +474,9 @@ public function admin_panel() {
 				'Yoast SEO',
 				'WooCommerce'
 			);
-			echo "</p>\n";
+			echo '</p>';
 
+			echo '<fieldset id="id-name" class="yoast-field-group">';
```
JoseeWouters, line 479: Why `id-name`? Where is this coming from?

classes/woocommerce-seo.php
JoseeWouters: Decided to leave this, because the class is probably not necessary anymore.

PR #626, opened 2020-07-20
Set new hidden columns defaults only if user did not change the previous defaults

afercia:
## Context

* We want to cover the case were users may had set their custom settings for the Yoast columns and then they activate SEO Woo: in this case the user custom settings should not be changed. This way, new "defaults" are set only if the current setting is the "old defaults".
For reference, see the original issue https://github.com/Yoast/featurerequests/issues/464 

## Summary

* Improves setting new defaults for the columns visibility in the Products overview page.

* checks if the current hidden columns user setting is the same as the old defaults, where these columns were the hidden ones:
- wpseo-title
- wpseo-metadesc
- wpseo-focuskw

## Test instructions

- activate Yoast SEO and WooCommerce
- go to the Products page
- set custom settings for the Yoast columns in the Screen Options 
- activate SEO Woo
- go to the Products page
- observe your columns settings **did not change**
- deactivate SEO Woo
- go back to the Products page
- set the columns setting to the old defaults: all visible except:
  - wpseo-title
  - wpseo-metadesc
  - wpseo-focuskw
- activate SEO Woo
- go back to the Products page
- observe that this time the settings did change to the new default: only SEO score is visible
- check the database: a user meta `wpseo_woo_columns_hidden_default` has been set for your user

Fixes [P3-80]

### DIFF
--- a/classes/woocommerce-seo.php
+++ b/classes/woocommerce-seo.php
@@ -551,10 +551,11 @@ class Yoast_WooCommerce_SEO {
 
 		$user_hidden_yoast_columns = array_filter( $user_hidden_columns, [ $this, 'filter_yoast_columns' ] );
 
-		$is_old_default = ( count( array_diff( $yoast_hidden_columns_old_defaults, $user_hidden_yoast_columns ) ) === 0 ) ? true : false;
+		$is_old_default = count( array_diff( $yoast_hidden_columns_old_defaults, $user_hidden_yoast_columns ) ) === 0;
 
 		// Don't do anything if the Yoast hidden columns old defaults have been changed by the user.
 		if ( ! $is_old_default ) {
+			update_user_option( $user_id, 'wpseo_woo_columns_hidden_default', '1', true );
 			return;
 		}
 
@@ -592,17 +593,6 @@ class Yoast_WooCommerce_SEO {
 	 */
 	private function filter_yoast_columns( $column ) {
 		return strpos( $column, 'wpseo-' ) === 0;
-	}
-
-	/**
-	 * Filter items if they have a count of zero.
-	 *
-	 * @param array $item The item to potentially filter out.
-	 *
-	 * @return bool Whether or not the count is zero.
-	 */
-	private function filter_items( $item ) {
-		return $item['count'] !== 0;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


classes/woocommerce-seo.php
```diff
@@ -536,17 +536,35 @@ public function footer_js() {
...
+			'wpseo-title',
+			'wpseo-metadesc',
+			'wpseo-focuskw',
+		];
+
 		$user_hidden_columns = get_hidden_columns( $current_screen );
 
+		$user_hidden_yoast_columns = array_filter( $user_hidden_columns, [ $this, 'filter_yoast_columns' ] );
+
+		$is_old_default = ( count( array_diff( $yoast_hidden_columns_old_defaults, $user_hidden_yoast_columns ) ) === 0 ) ? true : false;
```
herregroen, line 554: No need for the tertiary, `===` already returns a boolean.
```suggestion
		$is_old_default = ( count( array_diff( $yoast_hidden_columns_old_defaults, $user_hidden_yoast_columns ) ) === 0 );
```

classes/woocommerce-seo.php
```diff
@@ -536,17 +536,35 @@ public function footer_js() {
...
+
+		// Don't do anything if the Yoast hidden columns old defaults have been changed by the user.
+		if ( ! $is_old_default ) {
+			return;
```
herregroen, line 558: I would also set the user option here, if they've been changed they don't need to be reset later.

classes/woocommerce-seo.php
```diff
@@ -565,6 +583,28 @@ public function set_yoast_columns_hidden_by_default( $current_screen ) {
...
+	}
+
+	/**
+	 * Filter items if they have a count of zero.
+	 *
+	 * @param array $item The item to potentially filter out.
+	 *
+	 * @return bool Whether or not the count is zero.
+	 */
+	private function filter_items( $item ) {
```
herregroen, line 604: I see no other usages of this function. Why was this added?

afercia: Copy-pasta leftover 🍝 😛  (it's used in Yoast SEO)

afercia:
Thanks @herregroen! I think I addressed the pending points.

IreneStr:
I'll cherrypick this PR to the release branch.

PR #627, opened 2020-07-23
Travis: change from "trusty" to "xenial"

jrfnl:
## Context

* Improved CI / build testing

## Summary

* Improved CI / build testing

As the "trusty" environment is no longer officially supported by Travis, they decided in their wisdom to silently stop updating the PHP "nightly" image, which makes it next to useless as the last image apparently is from January....

This updates the Travis config to:
* Use the `xenial` distro, which at this time is the default.
* Makes the expected OS explicit (linux).
* The `xenial` distro is quite bare, so `mysql` is not available by default. Adding this as a `service` fixes that.

## Test instructions

* Verify that the build passes and that the third column on the Travis build page shows `xenial`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
+os: linux
+dist: xenial
 language: php
-dist: trusty
+
+services:
+  - mysql
 
 branches:
   only:


enricobattocchi:
Verify that the build passes and that the third column on the Travis build page shows `xenial` :heavy_check_mark:

PR #640, opened 2020-10-14
Fix grunt watch

moorscode:
## Context

* `grunt watch` is currently not working, this has recently been fixed in Yoast SEO and should work in every plugin.

## Summary

* Tooling: grunt watch (& build-watch) fix

* Moved the Grunt config to the config-structure as introduced in Yoast SEO
* Added ESLint checking to Travis
* Fixed ESLint errors

## Test instructions

* Run `yarn`
* Run `grunt watch`
* Modify JavaScript, PHP or Grunt-configuration files
* See the appropriate checks being performed for the changed file (linting/CS for PHP, ESLint and webpack build for JS, ESLint for grunt-files)

Also introduces `build-watch`

* Run `yarn`
* Run `grunt build-watch`
* See the composer install command being executed
* See the JavaScript files being build
* See the watcher being started

Fixes #


config/grunt/task-config/aliases.yaml
```diff
@@ -5,6 +5,11 @@ build:
   - 'build:i18n'
   - 'build:js'
 
+'build-watch':
+  - 'shell:composer-install'
+  - 'build:js'
```
igorschoester, line 10: Couple of things that are a bit strange to me. Though I'm not against this approach.

You need the `composer install` to run watch. However, when calling `watch` without the build this is not enforced. So you see it as a `build` step? But then it is not in the `grunt build` task.

You run `build:js`, not `build`. In this case the translations are skipped. Why?

igorschoester (changes requested):
CR ... a question, but not blocking.

As for acceptance, it works, but not useful for the PHP part at the moment. It does not find the nested PHP files, only the ones in the root. Changing the `files.php` to `**/*.php` would fix that. Though that might have other consequences?

config/grunt/task-config/aliases.yaml
moorscode: The build-watch is to make sure you start with a situation in which all the files are up-to-date that are required to run the plugin with.

The language files are required for publication but not for development.

The `build-watch` only has to be triggered once after checking out a new branch.
When cancelled and a new watch is needed, just running `grunt watch` is sufficient and quicker.

moorscode: I think adding `composer install` to the `grunt build` step would a sensible addition.
We've never had that in there for any repository yet, so had not considered putting it in here.

moorscode:
Sorry, missed your question.

As included the `classes/**` in the copy task for the artifact, all PHP files are copied in that directive.

I think your suggestion is a sensible one but doesn't seem to be needed.
The only occurrences of `<%= files.php %>` are in the `addtextdomain` and `checktextdomain` tools, which are superseded by the PHPCS checks.

moorscode:
Sorry, failed to search in the current PR for occurrences 🤦 
Of course it's needed now!

config/grunt/task-config/aliases.yaml
igorschoester: > The language files are required for publication but not for development.
Makes sense, thanks for explaining 😄 

> We've never had that in there for any repository yet, so had not considered putting it in here.
Out of scope for this I reckon. I'm not feeling adding it now only on this repository.

igorschoester:
Broadened the scope of `files.php` to include PHP files inside the `classes` folder. I was thinking general `**/*.php` first, but that includes all the tests and vendor files too. Thought that was a bit much.

Acceptance ✅

Issue #526, opened 2020-02-08
Sale price, regular price and priceValidUntil


galbaras:
### What did you expect to happen?
Sales normally have a validity period, so there should be a field for it in products/variations.

When a product is on sale, its schema markup should include both the regular price and the sale price, similar to Google's specifications for product feeds.

### What happened instead?
priceValidUntil was just removed from the plugin, and only a single price appears in the markeup.

Google Search Console issues warnings to sites since v12.5 due the removed attribute.

### How can we reproduce this behavior?
Set a sale price for a product and check the schema markup on the product page.

### Can you provide a link to a page which shows this issue?

https://dev.get-business-online.com/pacificsprings/product/urbane-spring-water-cooler/

### Technical info
* WordPress version: 5.3.2
* Yoast SEO version: 13.0
* Yoast SEO for WooCommerce version: 12.5.1

Flobaby:
Thanks for starting this thread. Yep, Google is giving me validation problems for priceValidUntil. Just to add my 2 cents. :)

cavalierlife:
+1 . Please undo the change made in the latest version. Two days ago I posted the following in the wp.org support forum. Will repost here:

Yesterday’s Yoast changelog includes: “Removes the priceValidUntil attribute, as it’s not sensible.”

_Today, my client gets an email from Google that says

Search Console has identified that your site is affected by 1 Products issues:

Top Warnings

Warnings are suggestions for improvement. Some warnings can affect your appearance on Search; some might be reclassified as errors in the future. The following warnings were found on your site:

Missing field “priceValidUntil”

We recommend that you fix these issues when possible to enable the best experience and coverage in Google Search._

So I’m thinking maybe Google thinks it was sensible. Waiting for my client to call me in a panic….

Djennez:
Hmm, it's a bug that we don't show the `priceValidUntil` on (scheduled) sale products. So I'll mark this as bug specifically for that case. It is, however, intentional to not show this value on every product page, as it makes no sense on products that have a fixed price.

galbaras:
Thanks.

I think the removal could have been documented better. Reference to "regular price" would have made sense to everyone, I'm sure, especially after sale prices get the attribute back.

Pcosta88:
Please inform the customer of conversation # 582990 when this conversation has been closed.

Pcosta88:
Please inform the customer of conversation # 583176 when this conversation has been closed.

rmarcano:
[...]

PR #651, opened 2020-11-10
Apply the wpseo_schema_product filter to make the Product type filterable

johannadevos:
## Context

* https://yoast.atlassian.net/browse/P2-504

## Summary

* Makes the Schema `Product` type filterable.

* I looked for all occurrences of the filter `woocommerce_structured_data_***` in the code and found these three:
    * `woocommerce_structured_data_product`.
    * `woocommerce_structured_data_type_for_page`.
    * `woocommerce_structured_data_review`.
* The `woocommerce_structured_data_product` filter triggers the `change_product` method, which changes the Schema `Product` type. I added a call to our `wpseo_schema_product` filter to make the `Product` type filterable.
* The `woocommerce_structured_data_type_for_page` filter triggers the `remove_woo_breadcrumbs` method, which removes the Woo breadcrumbs from the Schema output. I did not add a call to a Yoast filter here, since `remove_woo_breadcrumbs` only has the responsibility to remove the Woo breadcrumbs.
* The `woocommerce_structured_data_review` filter triggers the `change_reviewed_entity` method. However, [the review filter is applied](https://github.com/woocommerce/woocommerce/blob/02cf0dfaed5923513de0c88add597d1560c2cfd2/includes/class-wc-structured-data.php#L368) in a method that is [no longer used, but not deprecated either](https://github.com/woocommerce/woocommerce/pull/24278#discussion_r312598861). Because the WooCommerce developers mention that they might reintroduce this method, and if so we don't want our code to break, I added a call to our `wpseo_schema_review` filter. But because `woocommerce_structured_data_review` is currently never applied, this change is not testable now. 

**Test the `Product` type filter.**
* Open the page source of a WooCommerce product page in the browser.
* Look for the WooCommerce Schema output (tip: search for `"@type": "Product"`).
    * If you don't see the Schema output, this should be fixable by adding a price to your product.
* Make sure the following code is executed, for example by adding it to `functions.php` of your theme.

```
	add_filter( 'wpseo_schema_product', 'example_change_product' );
	function example_change_product( $data ) {
		$data['cats'] = 'exist';
		return $data;
	}
```

* Refresh the page source.
* See that `"cats": "exist"` has been added to your Product Schema, and that the rest of the Schema is unchanged.

*

## Impact check

* The WooCommerce Schema output.

## UI changes

## Documentation

## Quality assurance

Fixes https://github.com/Yoast/wpseo-woocommerce/issues/526

### DIFF
--- a/classes/woocommerce-schema.php
+++ b/classes/woocommerce-schema.php
@@ -132,6 +132,13 @@ class WPSEO_WooCommerce_Schema {
 
 		$this->data['review'][] = $data;
 
+		/**
+		 * Filter: 'wpseo_schema_review' - Allow changing the Review type.
+		 *
+		 * @api array $data The Schema Review data.
+		 */
+		$this->data = apply_filters( 'wpseo_schema_review', $this->data );
+
 		return [];
 	}
 
--- a/classes/woocommerce-schema.php
+++ b/classes/woocommerce-schema.php
@@ -8,7 +8,7 @@
 use Yoast\WP\SEO\Config\Schema_IDs;
 
 /**
- * Class WPSEO_WooCommerce_Schema
+ * Class WPSEO_WooCommerce_Schema.
  */
 class WPSEO_WooCommerce_Schema {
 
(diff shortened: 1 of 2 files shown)


andizer (approved):
Cr 👍

PR #654, opened 2020-11-25
Unit Tests: make the testsuite compatible with PHPUnit 9 and test against PHP 8

jrfnl:
## Context

* Preparation for compatibility with PHP 8

## Summary

* Preparation for compatibility with PHP 8

Similar to the other test update PRs, this PR will be easiest to review by looking at each commit individually.

### Composer: require Yoast/wp-test-utils

* Adds a dev dependency to the `yoast/wp-test-utils` package, which will also make the PHPUnit Polyfills available.
* As that package already requires and manages the installable versions for BrainMonkey and PHPUnit, remove these as explicit requirements from `require- dev` in favour of letting the WP Test Utils package manage the versions.

The install is done based on PHP 5.6 and includes updating the dependencies from PHPUnit itself within the constraints of maintaining compatibility with  PHP 5.6.

This new package adds the following features:
* Polyfills for various PHPUnit cross-version changes.
* Basic test cases for use by the plugins.

Refs:
* https://github.com/Yoast/wp-test-utils
* https://github.com/Yoast/PHPUnit-Polyfills/

### Tests: use the bootstrap file from WP Test Utils

... and remove everything which is already done within that file.

### Tests: switch over to the Yoast\WPTestUtils\BrainMonkey\YoastTestCase

All test  classes for the BrainMonkey based unit tests in the WooCommerceSEO test suite have a  WooCommerceSEO native `TestCase` as a parent class.

This switches the parent of that  `TestCase` from the PHPUnit native TestCase to the `Yoast\WPTestUtils \BrainMonkey\YoastTestCase`.

The `YoastTestCase`:
* Inherits PHPUnit cross- version compatibility from the PHPUnit Polyfills TestCase.
* Inherits the  BrainMonkey set up and teardown and Mockery expectation counting as assertions from the BrainMonkey TestCase.
* And natively adds a number of stubs for  commonly used WP functions.

This switch over includes:
* Removing the addition of  function stubs which are already added via the `YoastTestCase` class.
* Removing  the `tearDown()` method. This will be inherited from the `YoastTestCase` now.
* Renaming the `setUp()` method to `set_up()` in various test classes for PHPUnit  cross-version compatibility.

### Tests: switch over to the Yoast\WPTestUtils\BrainMonkey\TestCase for one specific test class

Three tests within the `Schema_Test` test class set expectations for the `wp_strip_all_tags()` method.

However, the `wp_strip_all_tags()` method is now stubbed by default in the `YoastTestCase` `set_up()` method, which in effects means that the calls to the `wp_strip_all_tags()` method are not seen by Mockery and the expectations would not be met.

To maintain the existing test behaviour:
* The parent `TestCase` for this test class is switched from the `Yoast\WP\Woocommerce\Tests\TestCase` to the `Yoast\WPTestUtils\BrainMonkey\TestCase` which doesn't contain the stubbing of the `wp_strip_all_tags()` method;
* The expectations for the `get_option*()` methods are copied from the `Yoast\WP\Woocommerce\Tests\TestCase` `set_up()` method to the `set_up()` method in this test class.class
* And the reason for using a different parent `TestCase` are documented in the `set_up()` method docblock.

### TestCase: remove the generic stubbing of the translation and escaping functions

Only a very limited number of test need these stubs, so we may as well only load them if and when needed.

Includes removing some stubbing being done in certain test functions for a limited set of these function, which were a) previously already covered in the test `setUp()` and b) are better served by using the dedicated functions for stubbing the complete set of functions.

### Test: remove in test stubs which are handled in the set up

A number of generic function stubs are now provided via the `set_up()` function in the base `TestCase`, so individual test functions don't need to duplicate these stubs.

### Tests: modernize used assertions

The PHPUnit Polyfills allows for using PHPUnit  9.x assertions, even when the tests are being run on older PHPUnit versions.

This updates select assertions to the modern PHPUnit syntax, most notably, it switches out the following assertions:
* `assertInternalType()` to `assertIs [Type]()`.
* `assertContains()` with string haystacks for `assertStringContainsString()`.

### Tests: remove use of `assertAttribute*()` assertions

The `assertAttribute*()` assertion methods were deprecated in PHPUnit 8.x and removed in PHPUnit 9.

The reasoning for that was that `protected` and `private` properties are implementation details and should not be tested directly, but via the methods in the class.

For now, I've elected to maintain the existing test behaviour, while making it work in a PHPUnit cross-version compatible manner.

This is done by:
1. Using a test helper method introduced provided via PHPUnit Polyfills (0.2.0) to get the value of a non-public property.
2. Changing the assertion for the tests from `assertAttribute*()` to `assert*()`

At a later point in time, it should be re-examined whether these implementation details should be tested in this way.

### Travis: run the unit tests against PHP 8/nightly

Now the BrainMonkey based tests are fully compatible with PHPUnit 9, the test suite can be run on PHP 8.

However, as the integration tests are not (yet) compatible with PHPUnit > 5.7, we need to add a separate condition block to the `script` section.

In this block, we run a `composer update` for just and only the test dependencies on PHP 8 to make sure that the test dependencies available are compatible with PHP 8.
And as the `composer.json` contains a hard-coded `platform.php` setting, we use `ignore-platform-reqs`  to get round that setting.

Includes:
* Adding `travis_retry` to `composer install`-like commands to get round builds stalling on the connection with Packagist, especially on older PHP versions.
* Adding the PHPUnit cache file which is generated when using PHPUnit 8/9 to `.gitignore`.

### PHPUnit config: minor tweaks

* Improve the code coverage configuration:
    - Add `addUncoveredFilesFromWhitelist` and  `processUncoveredFilesFromWhitelist` directives with sensible values.
    - Add the `forceCoversAnnotation` attribute to only record code coverage  for the method under test as annotated via a `@covers` annotation.

Based on this configuration, the strict code coverage for this testsuite is currently 48.28%.

* Verify the output of all builds for this PR, making sure that 1) the installation via Composer works without issues, 2) the tests are being run and 3) the tests pass.

To test locally, it will be simplest to:
1. Run `composer install`.
2. Run `composer test` on any PHP version below 8.0 and see the tests pass, including a more accurate assertion count (548 compared to 131 previously, due to Mockery expectations now being counted).
    This confirms that the changes to the `TestCase` works correctly cross-version as the PHPUnit version installed based on the `composer.lock` file will be PHPUnit 5.7.27.
3. Switch to PHP 8.
4. Run `composer update yoast/wp-test-utils --with-dependencies --ignore-platform-reqs`
5. Run `composer test` to see the tests running and passing on PHP 8 in combination with PHPUnit 9.x.
    This confirms that the tests can now run cross-version on all PHPUnit versions between PHPUnit 5.7 and PHPUnit 9.x and that the tests run and pass on PHP 8.

Note: this doesn't confirm full PHP 8 compatibility yet as the test code coverage base on the BrainMonkey based tests alone, as noted above, is only ~48%.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ phpcs.xml
 phpunit.xml
 yarn-error.log
 /js/dist/
+.phpunit.result.cache
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,12 +79,12 @@ before_install:
 
 install:
 - |
-  if [[ $TRAVIS_PHP_VERSION == "nightly" && "$PHPLINT" == "1" ]]; then
-    composer install --no-interaction --ignore-platform-reqs
+  if [[ $TRAVIS_PHP_VERSION == "nightly" ]]; then
+    travis_retry composer install --no-interaction --ignore-platform-reqs
   elif [[ "$PHPCS" == "1" || "$PHPUNIT" == "1" || "$PHPLINT" == "1" ]]; then
-    composer install --no-interaction
+    travis_retry composer install --no-interaction
   elif [[ "$TRAVIS_BUILD_STAGE_NAME" == "deploy" ]]; then
-    composer install --no-dev --no-interaction
+    travis_retry composer install --no-dev --no-interaction
   fi
 - if [[ "$CHECKJS" == "1" ]]; then yarn global add grunt-cli; yarn install; fi
 - if [[ "$SECURITY" == "1" ]]; then wget -P $SECURITYCHECK_DIR https://get.sensiolabs.org/security-checker.phar && chmod +x $SECURITYCHECK_DIR/security-checker.phar; fi
@@ -107,7 +107,7 @@ before_script:
   if [[ "$PHPUNIT" == "1" ]]; then
     git clone --depth=1 --branch="trunk" https://github.com/Yoast-dist/wordpress-seo.git $WP_DEVELOP_DIR/src/wp-content/plugins/wordpress-seo
     cd /tmp/wordpress/src/wp-content/plugins/wordpress-seo
-    composer install --no-dev --no-scripts --no-interaction --ignore-platform-reqs
+    travis_retry composer install --no-dev --no-scripts --no-interaction --ignore-platform-reqs
     cd -
   fi
 
@@ -151,6 +151,13 @@ script:
 - |
   if [[ "$PHPUNIT" == "1"  ]]; then
     travis_fold start "PHP.tests" && travis_time_start
+    composer test
+    travis_time_finish && travis_fold end "PHP.tests"
+  fi
+- |
+  if [[ $TRAVIS_PHP_VERSION == "nightly" ]]; then
+    travis_fold start "PHP.tests" && travis_time_start
+    travis_retry composer update yoast/wp-test-utils --with-dependencies --ignore-platform-reqs --no-interaction &&
     composer test
     travis_time_finish && travis_fold end "PHP.tests"
   fi
--- a/composer.json
+++ b/composer.json
@@ -36,10 +36,9 @@
     },
     "require-dev": {
         "yoast/yoastcs": "^2.1.0",
-        "phpunit/phpunit": "^5.7 || ^6.0 || ^7.0",
-        "brain/monkey": "^2.5",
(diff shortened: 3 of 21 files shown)


jrfnl:
Code coverage overview:
image

herregroen (approved):

PR #656, opened 2020-12-10
Integration Tests: make the testsuite x-version compatible and test against PHP 8

jrfnl:
## Context

* (Preparation for) compatibility with PHP 8

## Summary

* (Preparation for) compatibility with PHP 8

:bulb: This PR will be easiest to review by going through the individual commits one by one.

### Tests: remove redundant file includes

As there is a `autoload-dev` in place, the Composer autoloader will automatically include the files.

Includes removing a method which has now become redundant.

### Composer: update Yoast/wp-test-utils to 0.2.1

A new version of the WP Test Utils package has been released which deals with the necessary work-arounds for the WP integration tests and adds utilities for this which individual plugins can use, like functions for the bootstrap and a `TestCase` which includes all the polyfills from the PHPUnit Polyfill repo.

The install is done based on PHP 5.6 (`config.platform.php` is set in the `composer.json` file).

Refs:
* https://github.com/Yoast/wp-test-utils/releases/tag/0.2.0
* https://github.com/Yoast/wp-test-utils/releases/tag/0.2.1

### WPSEO_WooCommerce_UnitTestCase: switch parent class

This switches the parent class of the `WPSEO_WooCommerce_UnitTestCase` to the WP Test Utils `Yoast\WPTestUtils\WPIntegration\TestCase`.

Includes:
* Removing the `expectOutput()` method.
    This method is not used in this test suite and shouldn't be needed anyway as PHPUnit has build-in `expectOutputString()` and `expectOutputRegex()` methods which should be used instead.
    Additionally, WP Test Utils contains an `expectOutputContains()` method to complement the PHPUnit native methods.
* Removing other methods which were unused in this test suite.
    This applies to all the other methods.

This leaves the `WPSEO_WooCommerce_UnitTestCase` as an opaque class.
I've elected to let it remain though for the following reasons:
* Allows for adding future _plugin-specific_ test helper methods.
* Convenience as all test cases for this plugin `extend` from this one and any future changes in inheritance would only need to be applied to this class to apply them effectively to all test classes for this plugin.

Lastly, this commit makes sure that the generic test case is declared as `abstract` as it doesn't contain any tests itself.

### Tests: use the bootstrap utilities from WP Test Utils

For integration tests, WP Test Utils contains a `bootstrap-functions.php` file with utility functions for a number of common patterns used in bootstrap files for WP integration tests.

As the order in which things get loaded is _**extremely**_ important for the integration tests, whenever possible, the `WPIntegration\bootstrap_it()` method should be used.

This method will:
- Verify that the Composer `autoload.php` file exists.
- Load WordPress.
    The location for WP can be configured by setting either the `WP_TESTS_DIR` or the `WP_DEVELOP_DIR` environment variable on an OS level or via a custom `phpunit.xml` file.
    If neither of these two environment variables is found, it is presumed that the plugin is installed in `src/wp-content/plugins/plugin-name`.
    _This is compatible with the old bootstrap setup and developers will not need to make any changes to their environment for this to work._
- While loading WordPress, the plugin will be loaded, including the Composer autoload file.
- And lastly, a custom autoloader will be added to handle the PHPUnit 7.x MockObject classes not being compatible with PHP 8.0.
    This custom autoloader will load the copies of the PHPUnit 9.x MockObject classes which are part of the WP native test suite as of WP 5.6 when PHP 8.x is detected, and will let the composer autoloader load the PHPUnit native versions when not on PHP 8.x.

If anything goes wrong during these steps, the WP Test Utils method will provide human readable error messages, similar to before.

### PHPUnit config: improve code coverage configuration

* Add `addUncoveredFilesFromWhitelist` and `processUncoveredFilesFromWhitelist` directives with sensible values.
* Add the `forceCoversAnnotation` attribute to only record code coverage for the method under test as annotated via a `@covers` annotation.

Based on this configuration, the strict code coverage for this testsuite is currently 3.80%.

### Travis: run the tests against PHP 8.0

Now the test setup has been made compatible for allowing to run the tests on PHP 8 using PHPUnit 7.x, the integration tests can be run on PHP 8.0 in the CI runs.

However, there are still some things to take into account for running the tests on PHP 8.0:
1. The `composer.json` file contains a `config.platfom.php` setting which is set to PHP 5.6 and the repo has a committed `composer.lock` file which locks the PHPUnit version to PHPUnit 5.x.
2. WP hard limits their test suite to PHPUnit 7.x (max), which means the plugin integration tests need to be run on PHPUnit 7.x as well.
    However, PHPUnit 7.x is also _required_ for running the tests on PHP 8.0, PHPUnit 5.x won't work.
3. The BrainMonkey based tests, however, should be run on PHPUnit 9.3+, which is the first PHPUnit version officially compatible with PHP 8.0.

Other relevant information:
* Since this Friday, Travis now has a PHP 8.0 image available.
* Any PHP 8 related fixes have gone into WP 5.6. Running the integration tests against earlier WP versions is futile.

So, keeping all of that in mind, this commit:
* Adds a new build against PHP 8.0, with `WP_VERSION` set to WP `5.6` as released on Tuesday, and the `PHPUNIT` flag set to `1` (= on).
* In the `install` section: makes sure that the `composer install` with `--ignore-platform-reqs` is used for PHP 8.x as otherwise the install would fail due to PHPUnit 5.x (as per the lock file) not being allowed to be installed on PHP 8.
* In the `script` section:
    - Have a separate condition-based script for running the integration tests on PHP 8 (or `nightly`, though they currently won't be run against `nightly`).
    - Within that script **hard** require PHPUnit 7..x.
    - As the `PHPUNIT` flag has been added, the conditions for the BrainMonkey based unit tests are also adjusted.
        These tests are no longer run against `nightly`, but are run against `8.0` when the `PHPUNIT` flag is enabled.
    - Within the BrainMonkey PHP 8 script: remove the PHPUnit 7.x version which was required for the integration tests.
        This will then allow the `composer update` of the Test Utils (which was already in place) to install PHPUnit 9.x for running these tests.

Includes updating the matrix for the release of WP 5.6.

### Bootstrap: use a different pattern to load the plugins

This should hopefully prevent a double include of the Composer autoloader which led to fatal "Cannot redeclare composerRequire...." errors.

* Verify the output of all builds for this PR, making sure that 1) the installation via Composer works without issues, 2) the tests are being run, 3) the tests pass, 4) the BrainMonkey tests are being run and also pass.

To test locally, it will be simplest to:
1. Run `composer install`.
2. Run `composer integration-test` on any PHP version below 8.0 and see the tests pass.
    This confirms that the changes to the test bootstrap and the `TestCase` work correctly.
3. Switch to PHP 8.
4. Run `composer require --dev phpunit/phpunit:"^7.5" --update-with-dependencies --ignore-platform-reqs`
5. If the `WP_PLUGIN_DIR` environment variable on your system points to a directory containing a dev copy of WPSEO, run `composer install --no-dev --no-scripts  --ignore-platform-reqs` in the WPSEO directory to make sure that the PHPUnit version from WPSEO won't interfere.
6. Run `composer integration-test` to see the tests running and passing on PHP 8 in combination with PHPUnit 7.x.
    This confirms that the tests can now run cross-version on all PHPUnit versions between PHPUnit 5.7 and PHPUnit 7.x and that the tests run and pass on PHP 8.

Note: this doesn't confirm full PHP 8 compatibility yet as the test code coverage of the BrainMonkey + integration tests combined, is nowhere near 100%.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,17 +19,19 @@ jobs:
   fast_finish: true
   include:
     - php: 7.3.24
-      env: WP_VERSION=5.5 WP_MULTISITE=1 PHPUNIT=1 PHPCS=1 CHECKJS=1 SECURITY=1 TRAVIS_NODE_VERSION=lts/*
+      env: WP_VERSION=5.6 WP_MULTISITE=1 PHPUNIT=1 PHPCS=1 CHECKJS=1 SECURITY=1 TRAVIS_NODE_VERSION=lts/*
     - php: 5.6
-      env: WP_VERSION=5.4 PHPLINT=1 PHPUNIT=1
-    - php: 7.4
       env: WP_VERSION=5.5 PHPLINT=1 PHPUNIT=1
+    - php: 7.4
+      env: WP_VERSION=5.6 PHPLINT=1 PHPUNIT=1
     - php: 7.0
       env: WP_VERSION=master WP_MULTISITE=1 PHPUNIT=1
+    - php: 8.0
+      env: WP_VERSION=5.6 PHPLINT=1 PHPUNIT=1
     - php: "nightly"
       env: PHPLINT=1
     - stage: deploy
-      env: WP_VERSION=5.5
+      env: WP_VERSION=5.6
       if: tag IS present
       before_install:
         - openssl aes-256-cbc -K $encrypted_2b922af4d08d_key -iv $encrypted_2b922af4d08d_iv -in ./deploy_keys/wpseo_woo_deploy.enc -out ./deploy_keys/wpseo_woo_deploy -d
@@ -79,7 +81,7 @@ before_install:
 
 install:
 - |
-  if [[ $TRAVIS_PHP_VERSION == "nightly" ]]; then
+  if [[ ${TRAVIS_PHP_VERSION:0:1} == "8" || $TRAVIS_PHP_VERSION == "nightly" ]]; then
     travis_retry composer install --no-interaction --ignore-platform-reqs
   elif [[ "$PHPCS" == "1" || "$PHPUNIT" == "1" || "$PHPLINT" == "1" ]]; then
     travis_retry composer install --no-interaction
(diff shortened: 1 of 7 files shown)


jrfnl:
Code coverage info (integration tests only):
image

image

IreneStr (approved):
CR 👍

hansjovis:
Acceptance test: ✅

PR #661, opened 2021-01-19
Added schema presenter.

hansjovis:
## Context

* The old code used `WPSEO_Utils::schema_output`, which we deprecated.

## Summary

* Fixes a bug where a deprecation notice would be shown in the PHP debug log when visiting a product page.

* Added a new presenter, since...
	* the graph has already been generated and can be passed directly to the presenter.
	* the `Schema_Presenter` does not have the ability to add additional classes to the `script` tag.

* Starting with a vanilla installation: (debug logging enabled)
* Install, activate, and set up WooCommerce. (options below)
* Install and activate Yoast SEO (free or premium)
* Install and activate Yoast SEO WooCommerce
* Go to Admin > Products
* View a product
* Check the source code of the page, it should contain a `<script type="application/ld+json" class="yoast-schema-graph">` at the bottom of the page.
	* This script tag should contain a schema graph with one `Product` graph piece.
* Check the website's debug log, it should **not** contain the following deprecation warning:
  ```
  Deprecated:  WPSEO_Utils::schema_output is <strong>deprecated</strong> since version WPSEO 15.5 with no alternative available.
  ```

*

## Impact check

*

## UI changes

## Documentation

## Quality assurance

* [x] I have added unittests to verify the code works as intended

Fixes #

### DIFF
--- a/classes/presenters/woocommerce-schema-presenter.php
+++ b/classes/presenters/woocommerce-schema-presenter.php
@@ -1,0 +1,77 @@
+<?php
+/**
+ * WooCommerce Yoast SEO plugin file.
+ *
+ * @package WPSEO/WooCommerce
+ */
+
+use Yoast\WP\SEO\Presenters\Abstract_Indexable_Presenter;
+
+/**
+ * Presents the schema output inside a script tag.
+ */
+class WPSEO_WooCommerce_Schema_Presenter extends Abstract_Indexable_Presenter {
+
+	/**
+	 * The Schema to output.
+	 *
+	 * @var array
+	 */
+	protected $graph;
+
+	/**
+	 * The classes to add to the script tag.
+	 *
+	 * @var string[]
+	 */
+	protected $classes;
+
+	/**
+	 * WPSEO_WooCommerce_Schema_Presenter constructor.
+	 *
+	 * @param array    $graph   The schema graph.
+	 * @param string[] $classes The classes to add to the script tag.
+	 */
+	public function __construct( $graph, $classes ) {
+		$this->graph   = $graph;
+		$this->classes = $classes;
+	}
+
+	/**
+	 * Gets the raw schema value as an associative array.
+	 *
+	 * @return array The raw schema.
+	 */
+	public function get() {
+		return $this->graph;
+	}
+
+	/**
+	 * Presents the schema in a script tag.
+	 *
+	 * @return string The schema in a script tag.
+	 */
+	public function present() {
+		$graph = $this->get();
+
+		$schema = [
+			'@context' => 'https://schema.org',
+			'@graph'   => $graph,
+		];
+
+		$classes_string = \implode( ' ', $this->classes );
+
+		$output = \WPSEO_Utils::format_json_encode( $schema );
+		$output = \str_replace( "\n", \PHP_EOL . "\t", $output );
+		return '<script type="application/ld+json" class="' . \esc_attr( $classes_string ) . '">' . $output . '</script>';
+	}
+
+	/**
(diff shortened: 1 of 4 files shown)


classes/presenters/woocommerce-schema-presenter.php
```diff
@@ -0,0 +1,77 @@
...
+		$schema = [
+			'@context' => 'https://schema.org',
+			'@graph'   => $graph,
+		];
+
+		$classes_string = \implode( ' ', $this->classes );
+
+		$output = \WPSEO_Utils::format_json_encode( $schema );
+		$output = \str_replace( "\n", \PHP_EOL . "\t", $output );
+		return '<script type="application/ld+json" class="' . \esc_attr( $classes_string ) . '">' . $output . '</script>';
```
johannadevos, line 66: In the deprecated `schema_tag` function, the return statement includes `"\n"` at the very end. Did you leave this out on purpose?

classes/presenters/woocommerce-schema-presenter.php
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	/**
+	 * Returns the output as string.
+	 *
+	 * @return string The output.
+	 */
+	public function __toString() {
```
johannadevos, line 74: Should this method be included? Because your class extends the `Abstract_Indexable_Presenter`, and the `Abstract_Indexable_Presenter` extends the `Abstract_Presenter` which has the exact same `__toString()` method.

johannadevos:
CR: 🚧 

Overall the solution looks good and thoroughly implemented. I have two minor questions as comments.
Plus: I think you should add a unit test for the presenter you created :)

classes/presenters/woocommerce-schema-presenter.php
hansjovis: Good catch! I did not leave it out on purpose, just had not noticed that the original code added a `\n` to the end. I will add it.

classes/presenters/woocommerce-schema-presenter.php
hansjovis: You are right. However, I noticed that leaving this out made writing unit tests a lot more difficult, since I had to mock the method on the `Abstract_Presenter`.

In the end I opted for adding it to this class as well.

hansjovis:
> I think you should add a unit test for the presenter you created :)

Is it alright if I create a new issue for this? We are in a bit of a hurry, and the code is for a large part already tested by `tests/classes/schema-test.php`.

classes/presenters/woocommerce-schema-presenter.php
johannadevos: Ah 💡

johannadevos:
CR: 👍

johannadevos:
Acceptance: 🚧 

On `trunk`, the contents of the `@graph` are in an array (`[]`).
Screenshot 2021-01-20 at 13 43 45

On this branch, the array has disappeared:
Screenshot 2021-01-20 at 13 44 44

While I don't know the details about this, I'm assuming we want to keep the output exactly as it was before.

johannadevos:
Acceptance: 👍 

I didn't see the log on either branch, but since the call to `WPSEO_Utils::schema_output` has been [removed](https://github.com/Yoast/wpseo-woocommerce/pull/661/files#diff-c6aba86af0e18e74f7cd35c328c2c308c8e0aee0c188afbb7de98793e603f990L93) we know this problem must have been solved.

I did a [search](https://github.com/search?q=org%3AYoast+schema_output&type=code) in the Yoast organization on GitHub and found no other calls to `schema_output`.

PR #723, opened 2021-09-28
Update (fix) test setup

jrfnl:
## Context

* Update test setup

## Summary

* Updates test setup

See the [Make Core dev-note about the test changes](https://make.wordpress.org/core/2021/09/27/changes-to-the-wordpress-core-php-test-suite/) for the context of these changes.

### PHPUnit: update configuration

PHPUnit recently released version 9.5.10 and 8.5.21.

This contains a particular (IMO breaking) change:

> * PHPUnit no longer converts PHP deprecations to exceptions by default (configure `convertDeprecationsToExceptions="true"` to enable this)

Let's unpack this:

Previously (PHPUnit < 9.5.10/8.5.21), if PHPUnit would encounter a PHP native deprecation notice, it would:
1. Show a test which causes a deprecation notice to be thrown as **"errored"**,
2. Show the **first** deprecation notice it encountered and
3. PHPUnit would exit with a **non-0 exit code** (2), which will fail a CI build.

As of PHPUnit 9.5.10/8.5.21, if PHPUnit encounters a PHP native deprecation notice, it will no longer do so. Instead PHPUnit will:
1. Show a test which causes a PHP deprecation notice to be thrown as **"risky"**,
2. Show the **all** deprecation notices it encountered and
3. PHPUnit will exit with a **0 exit code**, which will show a CI build as passing.

This commit reverts PHPUnit to the previous behaviour by adding `convertDeprecationsToExceptions="true"` to the PHPUnit configuration.

Refs:
* https://github.com/sebastianbergmann/phpunit/blob/9.5/ChangeLog-8.5.md
* https://github.com/sebastianbergmann/phpunit/blob/9.5/ChangeLog-9.5.md

### Composer: update to WP Test Utils 1.0.0

WP Test Utils 1.0.0 provides cross-version compatibility with the test setup from WP < 5.9 and WP 5.9+. 

Includes updating the dependencies of WP Test Utils.

Refs:
* https://github.com/Yoast/wp-test-utils/releases/
* https://github.com/Yoast/PHPUnit-Polyfills/releases/
* https://github.com/mockery/mockery/releases/tag/1.3.5
* https://github.com/antecedent/patchwork/releases

### Travis: update for using correct PHPUnit version with WP 5.9+

As of WP 5.9, the integration tests can run on PHPUnit 5.7 - 9.x and should use the most appropriate PHPUnit version for the PHP version on which the tests are being run.

For WP < 5.9, the range of supported PHPUnit versions is still PHPUnit 5.7 - 7.x.

With that in mind and with there being a) a `config - platform - php` setting and b) a committed `composer.lock` file, some version juggling needs to be done.

For WP < 5.9, the logic previously used remains in place.
For WP 5.9, new logic is introduced.

As the conditions if they were to be made stand-alone would get pretty complicated, I've resorted to using `if - elif` conditions instead of stand-alone conditions.

And to make sure the various composer changes don't interfere with each other between the test runs for the integration tests and the Brainmonkey tests, a git reset of the `composer.lock` and the `composer.json` files and a `composer install` is run between the integration tests and the unit tests runs.

Note:
* To make sure the various composer changes don't interfere with each between the test runs for the integration tests and the brainmonkey tests, a git reset of the `composer.lock` and the `composer.json` files is done and a `composer install` is run at the end of select test runs.

### Travis: remove nightly

At this time, the current PHP 8.0 version is `8.0.11`. The Travis PHP `8.0` image yields PHP `8.0.9`, which is good enough for PHP 8.0.

`nightly` however is PHP `8.0.3` and hasn't been updated since Feb 2021, so running tests against that image is completely useless at this time.

Notes:
    * Tested `nightly` against all relevant `dist`s - `xenial`, `bionic` and `focal` and none have a more current image.
    * The [PHP Build project](https://github.com/php-build/php-build/tree/master/share/php-build/definitions), which Travis pulls its images from, _does_ have a PHP `8.1snapshot` image available.
        I've tested to see if that image is available on Travis, but unfortunately, no luck there either.

* See that the last build against `trunk` failed: https://app.travis-ci.com/github/Yoast/wpseo-woocommerce/builds/238618104
* See that it passes again with this PR.
    - Verify that builds against WP 5.9 / `master` are run against the most appropriate PHPUnit version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -148,20 +148,28 @@ script:
       composer check-cs
       travis_time_finish && travis_fold end "PHP.code-style"
     fi
-  # PHP Tests
+  # PHP Integration Tests
   - |
-    if [[ "$PHPUNIT" == "1" && ${TRAVIS_PHP_VERSION:0:1} != "8" && $TRAVIS_PHP_VERSION != "nightly" ]]; then
+    if [[ "$PHPUNIT" == "1" && "$WP_VERSION" != "latest" ]] && [[ "$WP_VERSION" == "5.9" || "${WP_VERSION:0:1}" == "6" || "$WP_VERSION" == "master" ]]; then
+      travis_fold start "PHP.integration-tests" && travis_time_start
+      composer config --unset platform.php &&
+      travis_retry composer update yoast/wp-test-utils --with-all-dependencies --no-interaction &&
+      vendor/bin/phpunit -c phpunit-integration.xml.dist
+      travis_time_finish && travis_fold end "PHP.integration-tests"
+    elif [[ "$PHPUNIT" == "1" && ${TRAVIS_PHP_VERSION:0:1} != "8" && $TRAVIS_PHP_VERSION != "nightly" ]]; then
       travis_fold start "PHP.tests" && travis_time_start
       composer integration-test
       travis_time_finish && travis_fold end "PHP.tests"
-    fi
-  - |
-    if [[ "$PHPUNIT" == "1" ]] && [[ ${TRAVIS_PHP_VERSION:0:1} == "8" || $TRAVIS_PHP_VERSION == "nightly" ]]; then
+    elif [[ "$PHPUNIT" == "1" ]] && [[ ${TRAVIS_PHP_VERSION:0:1} == "8" || $TRAVIS_PHP_VERSION == "nightly" ]]; then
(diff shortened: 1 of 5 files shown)


moorscode (approved):
Looks good to me!

PR #728, opened 2021-11-30
Prevent outputting empty product identifiers

jdevalk:
## Summary

* Fixes a bug where an empty product identifier value could be output.

* Set an empty value for a product identifier like ISBN or GTIN and see that it's not output.

## Quality assurance

* [x] I have added unittests to verify the code works as intended

Fixes #IM-1612


karlijnbok:
CR & Acceptance: 👍

PR #740, opened 2022-01-11
CI: switch to GitHub Actions

jrfnl:
## Context

* CI/Quality control

## Summary

* CI/QA is now run via GitHub Actions

👉🏻 This PR will be easiest to review by reviewing each commit individually.

👉🏻 This PR (ab)uses the `feature/` branch prefix on purpose to test the `deploy` workflow. Once this PR has been reviewed and merged, the `feature` branch for this PR should be removed from the `dist` repo.

⚠️ ~~The `deploy` workflow is currently failing on the final step to `push` to the `dist` repo. This is a permission issue which we didn't encounter with the Test Helper/Duplicate Post workflows as those are `public` repos on `dist`, while this repo is `private` on `dist`.~~
~~Most likely cause is that the `YOASTBOT_CI_PAT_DIST` access token has not (yet) been added to this repo. Someone with admin access needs to verify this.~~
~~Both the `YOASTBOT_CI_GH_ACTIONS_TOKEN` as the  `YOASTBOT_CI_PAT_DIST` secrets should exist.~~
* ~~The `YOASTBOT_CI_GH_ACTIONS_TOKEN` secret should be the access token for the `dev` organisation for the VCS download of the Free repo from the `dev` repo.~~
* ~~The `YOASTBOT_CI_PAT_DIST` secret should be the access token for the `dist` organisation to allow the git push.~~

**Update**: The tokens for the `deploy` workflow have been fixed.

### Composer: require the composer/installers package

As the YoastSEO WooCommerce plugin is a wordpress plugin and should be installed as such, the `composer/installers` package should be a direct requirement, instead of relying on inheriting the requirement from YoastSEO, which is a `dev` dependency.

### CI: switch to GitHub Actions - step 1: code style

This commit:
* Adds a GH Actions workflow for the CI code style check and to validate the `composer.json` file.
* Removes those actions from the `.travis.yml` configuration.
* Adds a small tweak to the CS ruleset.

Notes:
1. Builds will run on all pushes and on pull requests, except pushes to `master`.
2. Builds can also be manually triggered.
    Note: manual triggering of builds has to be [explicitly allowed](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/). This is not a feature which is enabled by default.
3. If a previous GH actions run for the same branch hadn't finished yet when the same branch is pushed again, the previous run will be cancelled.
    In Travis, this was an option on the "Settings" page - "Auto cancellation" -, which was turned on for most, if not all, Yoast repos. The `concurrency` configuration in the GHA script emulates the same behaviour.
4. This package can occasionally run into a particular Git bug - `BUG: refs/files-backend.c:2992: initial ref transaction called with existing refs`.
    Extensive debugging has determined that this is _most likely_ due to a combination of 1) the use of VCS for YoastSEO, 2) the use of the `composer-installers` Composer plugin and 3) the use of an overload on the `installer-paths` for the Composer installer plugin for the YoastSEO download.
    This bug should probably at some point be reported to Composer, but the bug only crops up intermittently, not consistently, so it is _hard_ to create a reliable reproduction test case to illustrate the behaviour/bug.
    As a work-around, removing the problematic settings when they are not needed, prevents the workflow from running into the bug.
    Note: I've favoured removing the problem settings over removing the YoastSEO _requirement_. Keeping the _requirement_ will allow for the Composer download cache to be re-used in more cases, making the builds faster all round.
    Also note that as YoastSEO will now be installed into the `wp-content/plugins/` directory when using this workflow, this directory needs to be excluded from the CS scan.
5. Composer dependency downloads will be cached for faster builds using a [predefined GH action](https://github.com/marketplace/actions/install-composer-dependencies) specifically created for this purpose.
    The alternative would be to handle the caching manually, which would add three extra steps to the script.
    Note: Caching works differently between Travis and GH Actions.
    On GH Actions, once a cache has been created, it can't be updated. It can only be replaced by a new cache with a different key.
    As the PHP version, the `composer.json` and a potential `composer.lock` hash are all part of the key used by the above mentioned action, this difference should not have a significant impact.
    Ref: https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows
6. The CS check will display the results in the actions script output log, as well as display any violations found inline in the GitHub code view using the [cs2pr](https://github.com/staabm/annotate-pull-request-from-checkstyle) tool.

 Differences with the Travis implementation:
* This check will now run on ALL pushes and pulls.
    The branch filtering which was previously being applied in Travis has not been implemented for this script.
* The PHPCS check will now run against PHP 7.4 (previously 7.3.24).
* The `composer validate` command will now only be run against the PHP version used in the `cs` script (PHP 7.4), not against multiple PHP versions, but as the `composer-install` action does a (non-strict) `composer validate` since v2 under the hood, this shouldn't be an issue.

### CI: switch to GitHub Actions - step 2: security check

This commit:
* Adds a GH Actions workflow for the `composer.lock` dependency security check.
* Removes all references to that action from the `.travis.yml` configuration.

Notes:
1. Builds will run on all pushes and on pull requests.
2. In addition to this, this workflow will run automatically every Monday at 6am against the default branch via a cron job.
    This is especially relevant for repositories which are not actively receiving PRs every week.
    Notes about workflows with cron jobs:
    1. If a repository is forked, workflows will also run in the fork. This means that with a cron job, those will also be run on forks.
        As that is not the intention, I've added a condition to prevent the cron job from running on forks, while still allowing the workflow to run on forks for other events.
        Note: this only applies to pre-existing forks. For new forks, scheduled jobs are disabled by default.
    2. There is one big downside to using cron jobs in workflows and that is that those workflows will **_automatically be disabled after 60 days of inactivity in a repo_**.
        In this context, an "inactive" repo means that there have been no commits to the repo within the last 60 days.
        The repo owner (organisation owner) will receive an email notification a few days before this is about to happen and can prevent the workflow from being disabled, but that does mean that for repos with low activity, this workflow will need to be kept active by acting on the email once every two months.
        If the workflow would still happen to get disabled, it can be re-enabled by anyone with commit access on the "Actions" -> "Security Check" page of the repo.
        Refs:
        * https://docs.github.com/en/actions/managing-workflow-runs/disabling-and-enabling-a-workflow
        * https://github.community/t/do-disabled-scheduled-workflows-re-activate-on-repository-activity/160658
3. Builds can also be manually triggered.
    Note: manual triggering of builds has to be [explicitly allowed](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/). This is not a feature which is enabled by default.
4. If a previous GH actions run for the same branch hadn't finished yet when the same branch is pushed again, the previous run will be cancelled.
    In Travis, this was an option on the "Settings" page - "Auto cancellation" -, which was turned on for most, if not all, Yoast repos. The `concurrency` configuration in the GHA script emulates the same behaviour.
5. The security check is run via a predefined action from the same maintainers as the security check tool which was previously used.
    In the documentation, it is suggested to potentially cache the vulnerabilities database, but knowing how caching works on GHA, I'm not so sure that's a good idea as in that case, chances of checking against an outdated database are high.
    Ref: https://github.com/marketplace/actions/the-php-security-checker

Differences with the Travis implementation:
* This check will now run on ALL pushes and pulls.
    The branch filtering which was previously being applied in Travis has not been implemented for this script.
* The check will now also run via the cron job.
* No need to keep track of the current version number of the security check tool anymore as the predefined action will handle that.
* As this check is now run in a separate workflow, there is no risk that the security check will run against a `composer.lock` file which has been updated during previous steps in the script. It will always run against the `composer.lock` file **_as committed into the repo_**, making the check much more reliable compared to how this check was run previously via Travis.

### CI: switch to GitHub Actions - step 3: linting

This commit:
* Adds a GH Actions workflow for the PHP lint CI check.
* Removes all references to that action from the `.travis.yml` configuration, as well as does some additional clean up of parts of the Travis script which have now become unused.
* Adds a small tweak to the `lint` script in the `composer.json`.

Notes:
1. Builds will run on:
    - Select pushes using branch filtering similar to before.
    - All pull requests.
    - When manually triggered.
        Note: manual triggering of builds has to be [explicitly allowed](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/). This is not a feature which is enabled by default.
2. If a previous GH actions run for the same branch hadn't finished yet when the same branch is pushed again, the previous run will be cancelled.
    In Travis, this was an option on the "Settings" page - "Auto cancellation" -, which was turned on for most, if not all, Yoast repos. The `concurrency` configuration in the GHA script emulates the same behaviour.
3. The default ini settings used by the `setup-php` action are based on the `php.ini-production` configuration.
    This means that `error_reporting=E_ALL & ~E_DEPRECATED & ~E_STRICT`, `display_errors` is set to `Off` and `zend.assertions` is set to `-1` (= do not compile).
    For the purposes of CI, especially for linting and testing code, I'd recommend running with `zend.assertions=-1`, `error_reporting=-1` and `display_errors=On` to ensure **all** PHP notices are shown.
    Note: the defaults will be changed in the next major of `setup-php` to be based on the `php.ini-develop` configuration, but that may still be a while off.
    Refs:
    * shivammathur/setup-php#450
    * shivammathur/setup-php#469
4. While the `setup-php` action offers the ability to [install the PHAR file for Parallel Lint](https://github.com/shivammathur/setup-php#wrench-tools-support), I've elected not to use that option as it would mean that we would not be able to use the `composer lint` script in the workflow, which would mean that the CLI arguments would have to be duplicated between the `composer.json` file and the `lint.yml` file.
    IMO, that would make this a typical point of failure where updates would be done in one, but not the other.
    If, at some point in the future, the Parallel Lint tool would start to support a config file for the CLI arguments, removing this point of failure, this choice can be (and should be) revisited.
5. This package can occasionally run into a particular Git bug - `BUG: refs/files-backend.c:2992: initial ref transaction called with existing refs`.
    Extensive debugging has determined that this is _most likely_ due to a combination of 1) the use of VCS for YoastSEO, 2) the use of the `composer-installers` Composer plugin and 3) the use of an overload on the `installer-paths` for the Composer installer plugin for the YoastSEO download.
    This bug should probably at some point be reported to Composer, but the bug only crops up intermittently, not consistently, so it is _hard_ to create a reliable reproduction test case to illustrate the behaviour/bug.
    As a work-around, removing the problematic settings when they are not needed, prevents the workflow from running into the bug.
    Note: I've favoured removing the problem settings over removing the YoastSEO _requirement_. Keeping the _requirement_ will allow for the Composer download cache to be re-used in more cases, making the builds faster all round.
    Also note that as YoastSEO will now be installed into the `wp-content/plugins/` directory when using this workflow, this directory needs to be excluded from the linting.
6. Composer dependency downloads will be cached for faster builds using a [predefined GH action](https://github.com/marketplace/actions/install-composer-dependencies) specifically created for this purpose.
    The alternative would be to handle the caching manually, which would add three extra steps to the script.
    Note: Caching works differently between Travis and GH Actions.
    On GH Actions, once a cache has been created, it can't be updated. It can only be replaced by a new cache with a different key.
    As the PHP version, the `composer.json` and a potential `composer.lock` hash are all part of the key used by the above mentioned action, this difference should not have a significant impact.
    Ref: https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows
7. The Linting check will display the results inline in the GitHub code view using the [cs2pr](https://github.com/staabm/annotate-pull-request-from-checkstyle) tool.

Differences with the Travis implementation:
* There is a minor difference in the branch filtering being done for `push` events.
    Travis accepted PCRE regexes for the filtering. GH Actions uses glob patterns. The glob patterns now in place match the regexes as closely as possible, but are not an exact match to the "old" patterns. They should be sufficient for our purposes though.
* The tag-based branch filter has not been added as that filter was only intended for use with the `deploy` stage.
* Linting will now also be executed against PHP 8.1 and 8.2 (nightly).
    Linting runs against PHP 8.2 will be "allowed to fail" for the time being.
    Note: if any of the "allowed to fail" jobs actually fail, the workflow will show as successfull, but there will still be a red `x` next to a PR.
    This is a known issue in GHA: actions/toolkit#399
    There are work-arounds possible for this, however, the work-arounds would hide failures even more, meaning that chances are they won't be seen until they actually become a problem (no longer allowed to fail), which is too late.

### CI: switch to GitHub Actions - step 4: javascript checks

This commit:
* Adds a GH Actions workflow for the Javascript related CI checks.
* Removes all references to that check from the `.travis.yml` configuration.

Notes:
1. Builds will run on:
    - Select pushes using branch filtering similar to before.
    - All pull requests.
    - When manually triggered.
        Note: manual triggering of builds has to be [explicitly allowed](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/). This is not a feature which is enabled by default.
2. If a previous GH actions run for the same branch hadn't finished yet when the same branch is pushed again, the previous run will be cancelled.
    In Travis, this was an option on the "Settings" page - "Auto cancellation" -, which was turned on for most, if not all, Yoast repos. The `concurrency` configuration in the GHA script emulates the same behaviour.
3. The standard Ubuntu 20 image used by this action comes preloaded with Node, NPM and Yarn, so - in contrast to Travis - those don't need to be set up explicitly via script commands.
    Ref: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md
4. The build still uses the `setup-node` action as it allows for caching of the yarn dependencies for faster builds.
    The `node-version` key being passed, fixes the Node version to version 14, same as previously in Travis.
    Ref: https://github.com/actions/setup-node
5. Includes a "debug" step which will show the versions of the underlying tooling used, which should come in handy when the build would fail unexpectedly.

Differences with the Travis implementation:
* There are minor differences in the NPM and Yarn versions used.
    This has been discussed and been deemed acceptable.
* In Travis, the `install` step ran a `composer install`, including for the JS check.
    I have determined that this step is not actually needed for the JS check, so it is not included in this script.

Additional observations:
* The `yarn install` yields five warnings. These were also previously thrown in Travis. These should be looked into separately.
    ```
    warning "@yoast/grunt-plugin-tasks > babel-eslint@10.0.3" has unmet peer dependency "eslint@>= 4.12.1".
    warning "@yoast/grunt-plugin-tasks > eslint-plugin-jsx-a11y@6.2.3" has unmet peer dependency "eslint@^3 || ^4 || ^5 || ^6".
    warning "@yoast/grunt-plugin-tasks > eslint-plugin-react@7.17.0" has unmet peer dependency "eslint@^3.0.0 || ^4.0.0 || ^5.0.0 || ^6.0.0".
    warning "@yoast/grunt-plugin-tasks > eslint-plugin-jsx-a11y > axobject-query@2.1.1" has unmet peer dependency "eslint@^5 || ^6".
    warning "@yoast/grunt-plugin-tasks > eslint-plugin-react > eslint-plugin-eslint-plugin@2.1.0" has unmet peer dependency "eslint@>=5.0.0".
    warning " > grunt-git@1.0.14" has unmet peer dependency "grunt@~1.0.1".
    ```
* The `grunt eslint:plugin` task yields one system warning and one code warning. Again, this should be looked into separately:
    ```
    Warning: React version not specified in eslint-plugin-react settings. See https://github.com/yannickcr/eslint-plugin-react#configuration .

    /home/runner/work/wpseo-woocommerce/wpseo-woocommerce/js/src/assessments/ProductDescriptionAssessment.js
    Warning:   59:25  warning  Method 'scoreProductDescription' has a complexity of 7  complexity
    ```

### CI: switch to GitHub Actions - step 5: PHP unit tests

This commit:
* Adds a GH Actions workflow for the unit test CI check.
* Removes all references to that check from the `.travis.yml` configuration.

Notes:
1. Builds will run on:
    - Select pushes using branch filtering similar to before.
    - All pull requests.
    - When manually triggered.
        Note: manual triggering of builds has to be [explicitly allowed](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/). This is not a feature which is enabled by default.
2. If a previous GH actions run for the same branch hadn't finished yet when the same branch is pushed again, the previous run will be cancelled.
    In Travis, this was an option on the "Settings" page - "Auto cancellation" -, which was turned on for most, if not all, Yoast repos. The `concurrency` configuration in the GHA script emulates the same behaviour.
3. The default ini settings used by the `setup-php` action are based on the `php.ini-production` configuration.
    This means that `error_reporting=E_ALL & ~E_DEPRECATED & ~E_STRICT`, `display_errors` is set to `Off` and `zend.assertions` is set to `-1` (= do not compile).
    For the purposes of CI, especially for linting and testing code, I'd recommend running with `zend.assertions=-1`, `error_reporting=-1` and `display_errors=On` to ensure **all** PHP notices are shown.
    Note: the defaults will be changed in the next major of `setup-php` to be based on the `php.ini-develop` configuration, but that may still be a while off.
    Refs:
    * shivammathur/setup-php#450
    * shivammathur/setup-php#469
4. While the `setup-php` action offers the ability to [install the PHAR file for PHPUnit](https://github.com/shivammathur/setup-php#wrench-tools-support), I've elected not to use that option as we need to do a `composer install` anyway to get the other dependencies, like WP Test Utils.
5. This package can occasionally run into a particular Git bug - `BUG: refs/files-backend.c:2992: initial ref transaction called with existing refs`.
    Extensive debugging has determined that this is _most likely_ due to a combination of 1) the use of VCS for YoastSEO, 2) the use of the `composer-installers` Composer plugin and 3) the use of an overload on the `installer-paths` for the Composer installer plugin for the YoastSEO download.
    This bug should probably at some point be reported to Composer, but the bug only crops up intermittently, not consistently, so it is _hard_ to create a reliable reproduction test case to illustrate the behaviour/bug.
    As a work-around, removing the problematic settings when they are not needed, prevents the workflow from running into the bug.
    Note: I've favoured removing the problem settings over removing the YoastSEO _requirement_. Keeping the _requirement_ will allow for the Composer download cache to be re-used in more cases, making the builds faster all round.
6. Composer dependency downloads will be cached for faster builds using a [predefined GH action](https://github.com/marketplace/actions/install-composer-dependencies) specifically created for this purpose.
    The alternative would be to handle the caching manually, which would add three extra steps to the script.
    Note: Caching works differently between Travis and GH Actions.
    On GH Actions, once a cache has been created, it can't be updated. It can only be replaced by a new cache with a different key.
    As the PHP version, the `composer.json` and a potential `composer.lock` hash are all part of the key used by the above mentioned action, this difference should not have a significant impact.
    Ref: https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows
7. For these BrainMonkey based unit tests, we can (and should) use the most appropriate PHPUnit/BrainMonkey/Mockery/etc version for the PHP version the tests are being run on. Most notably, this is actually needed to get the tests running on PHP 8.0 and higher.
    As there is a committed `composer.lock` file and the PHPUnit version is locked at PHPUnit 5.x, this means we have to do an on-the-fly update of the PHPUnit version depending on the PHP version of the current build.
    As we still also want to benefit from the Composer caching which the `ramsey/composer-install` action sets up, I've done some nifty trickery with the options passed to the `composer-install` action to get that working.
    - The `dependency-versions: "highest"` basically changes the command used by the action from `composer install` to `composer update`, however we don't want to update _all_ dependencies as run-time dependencies should remain locked at their original version.
    - To that end, I'm passing additional `composer-options` which will limit the update to just and only the test dependencies.
    - That combined with the earlier `--unset platform.php` will ensure the most appropriate PHP version is used.

Differences with the Travis implementation:
* The most appropriate version of the test dependencies will now be used for _all_ builds, not just for builds again PHP 8.0/`nightly`.
* In addition to the PHP versions against which the tests were previously run on Travis, the tests will now also be run against PHP 8.1.
    As PHP 8.1 has been released and the build against PHP 8.1 is currently passing, this build is **_not_** set to be "allowed to fail".

### CI: switch to GitHub Actions - step 6: PHP integration tests

This commit:
* Adds a GH Actions workflow for the Integration test CI check.
* Removes all references to that check from the `.travis.yml` configuration.

Notes:
1. Builds will run on:
    - Select pushes using branch filtering similar to before.
    - All pull requests.
    - When manually triggered.
        Note: manual triggering of builds has to be [explicitly allowed](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/). This is not a feature which is enabled by default.
2. If a previous GH actions run for the same branch hadn't finished yet when the same branch is pushed again, the previous run will be cancelled.
    In Travis, this was an option on the "Settings" page - "Auto cancellation" -, which was turned on for most, if not all, Yoast repos. The `concurrency` configuration in the GHA script emulates the same behaviour.
3. In contrast to various other scripts, the matrix for these integration tests is explicitly set with all variables explicitly defined for each build, as we want full control over the three variables (`php_version`, `wp_version` and `multisite`) used for each run.
4. As a MySQL database is needed, we need to explicitly add that service.
    The MySQL version used varies based on the PHP version against which the tests are run for two reasons:
    1. It's good practice to test against multiple different MySQL version to verify that any queries run are MySQL cross-version compatible.
    2. While WP is _supposed to_ support MySQL 8.0 since WP 5.4, in actual fact, WordPress doesn't fully support MySQL 8.0 on PHP < 7.4.
        Also see: https://core.trac.wordpress.org/ticket/52496
5. The default ini settings used by the `setup-php` action are based on the `php.ini-production` configuration.
    This means that `error_reporting=E_ALL & ~E_DEPRECATED & ~E_STRICT`, `display_errors` is set to `Off` and `zend.assertions` is set to `-1` (= do not compile).
    For the purposes of CI, especially for linting and testing code, I'd recommend running with `zend.assertions=-1`, `error_reporting=-1` and `display_errors=On` to ensure **all** PHP notices are shown.
    Note: the defaults will be changed in the next major of `setup-php` to be based on the `php.ini-develop` configuration, but that may still be a while off.
    Refs:
    * shivammathur/setup-php#450
    * shivammathur/setup-php#469
6. While the `setup-php` action offers the ability to [install the PHAR file for PHPUnit](https://github.com/shivammathur/setup-php#wrench-tools-support), I've elected not to use that option as we need to do a `composer install` anyway to get the other dependencies, like WP Test Utils.
7. The PHPUnit version(s) supported and needed to run the tests depend on **both** the WP version used as well as the PHP version used.
    As an `if` condition on a step would get pretty complicated to handle this, I've introduced a separate -fully documented - step to determine the "type" of install we need.
    The output of that step is subsequently used in the `if` condition for follow-on steps to make sure the correct PHPUnit version will be installed by Composer.
8. Composer dependency downloads will be cached for faster builds using a [predefined GH action](https://github.com/marketplace/actions/install-composer-dependencies) specifically created for this purpose.
    The alternative would be to handle the caching manually, which would add three extra steps to the script.
    Note: Caching works differently between Travis and GH Actions.
    On GH Actions, once a cache has been created, it can't be updated. It can only be replaced by a new cache with a different key.
    As the PHP version, the `composer.json` and a potential `composer.lock` hash are all part of the key used by the above mentioned action, this difference should not have a significant impact.
    Ref: https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows
9. As there is a committed `composer.lock` file and the PHPUnit version is locked at PHPUnit 5.x, we have to do an on-the-fly update of the PHPUnit version depending on the WP and PHP version used for the current build.
    As we still also want to benefit from the Composer caching which the `ramsey/composer-install` action sets up, I've done some nifty trickery with the options passed to the `composer-install` action to get that working.
    - The `dependency-versions: "highest"` basically changes the command used by the action from `composer install` to `composer update`, however we don't want to update _all_ dependencies as run-time dependencies should remain locked at their original version.
    - To that end, I'm passing additional `composer-options` which will limit the update to just and only the test dependencies.
    - These type of updates are only run when needed based on the "install type" determination as described in (7).
10. The YoastSEO dependency will always be updated to the latest version available in `Yoast-dist` before running the tests.
    Note: The YoastSEO dependency is installed from the `Yoast-dist` repo for two reasons:
    1. This prevents a **_lot_** of extra steps being needed to get the YoastSEO `vendor_prefixed` directory in place, including juggling with the PHP version used for a test run. The version available from `Yoast-dist` already has the `vendor_prefixed` directory and the autoload file we need to run the integration tests, so this simplifies things immensely.
    2. The use of the VCS setting for the YoastSEO download would at random points in time, lead to errors during the `composer install` step as the package would run into a particular Git bug - `BUG: refs/files-backend.c:2992: initial ref transaction called with existing refs`.
        Extensive debugging has determined that this is _most likely_ due to a combination of 1) the use of VCS for YoastSEO, 2) the use of the `composer-installers` Composer plugin and 3) the use of an overload on the `installer-paths` for the Composer installer plugin for the YoastSEO download.
        Using the `Yoast-dist` repo instead of the VCS prevents this workflow from running into the bug.
11. Installing WordPress was previously done via inline commands in the Travis script. This has now been replaced by the more comprehensive `tests/bin/install-wp-tests.sh` script as made available by the WP CLI `scaffold` command.
    As WPCLI is not used anywhere else in the workflow, a copy of the script, with source annotations, is being introduced into the plugin codebase.
    Note: this script needs to have the `x` flag set to allow it to be executed.
    Ref: https://github.com/wp-cli/scaffold-command/blob/master/templates/install-wp-tests.sh
    Some differences between the scripts:
    - The plugin, nor the YoastSEO plugin, will no longer be copied into the WordPress install, as, to be honest, that's just not necessary with the current test setup/test bootstrap.
    - The location to which WP is downloaded and installed is different between the scripts, however, this makes no difference for our purposes as the WP Test Utils library already takes the default install locations from the `install-wp-tests.sh` script into account.
    - To test against the "nightly" of WordPress, the matrix needs to set the `wp_version` to `trunk` or `nightly` instead of `master` (as was previously used).
    - WordPress will be set up to always use `mysqli` for database commands, even on PHP 5.6.

Differences with the Travis implementation:
* The PHP/WP version matrix is largely the same as the one which was used in Travis, though some minor tweaks have been made.
* In addition to the PHP versions against which the tests were previously run on Travis, the tests will now also be run against PHP 8.1.
    As PHP 8.1 has been released and the build against PHP 8.1 is currently passing, this build is **_not_** set to be "allowed to fail" based on the PHP version, though as it uses (needs) WP `trunk`, it still, effectively, is allowed to fail until WP 5.9 has been released.
    Note: if a "allowed to fail" job actually fails, the workflow will show as successful, but there will still be a red `x` next to a PR.
    This is a known issue in GHA: actions/toolkit#399
    There are work-arounds possible for this, however, the work-arounds would hide failures even more, meaning that chances are they won't be seen until they actually become a problem (no longer allowed to fail), which is too late.
* Instead of using Git to clone WP, the WP-CLI script is now used to retrieve the relevant parts of WP and initiate the database and the `wp-config` settings.
* Instead of cloning a _second_ copy of YoastSEO (from `Yoast-dist`) and placing it within the WP `plugins` directory, the updated version of YoastSEO from `Yoast-dist` in the `vendor` directory is now used with the tests.
    Passing the `WP_PLUGIN_DIR` environment variable for the test runs makes sure that WP Test Utils sets this up correctly.
* In Travis, test runs would run either as single site or as multisite. In GH Actions, each build will always run the tests against WP in single site mode and for those builds with the `multisite` flag set to `true` in the matrix, the test will **_also_** be run in multisite mode.

### CI: switch to GitHub Actions - step 7: Deploy to dist

This commit:
* Adds a GH Actions workflow for deploying to the dist repo
* Removes the, now redundant, `.travis.yml` configuration.
* Removes the, now redundant, `scripts/deploy_to_dist.sh` script.
* Removes the, now redundant, `deploy_keys/wpseo_woo_deploy.enc` file.
* Updates the `.gitattributes` file to reflect the removal of these files.

Notes:
1. Builds will run on:
    - Select pushes using branch filtering similar to before.
        Note: this branch has been set up to match one of the filters to allow for testing the deploy.
    - Pushed tags.
    - When manually triggered.
        Note: manual triggering of builds has to be [explicitly allowed](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/). This is not a feature which is enabled by default.
2. If a previous GH actions run for the same branch hadn't finished yet when the same branch is pushed again, the previous run will be cancelled.
    In Travis, this was an option on the "Settings" page - "Auto cancellation" -, which was turned on for most, if not all, Yoast repos. The `concurrency` configuration in the GHA script emulates the same behaviour.
3. The workflow global `env` key contains information about the Yoast dist repo. This information is used as "variables" in select places in the rest of the script.
    **Important**: while the _default branch_ for the _development_ repo may be different, the `DIST_DEFAULT_BRANCH` env variable should reflect the _default branch_ for the _dist_ repo.
    Also note that the script presumes that the sister-repo in the `Yoast-dist` organisation will be called the same as the repo in the `Yoast` organisation. If ever this would not be the case, an additional env variable could be set here and used later on in the script to handle that.
4. The deployment is split into two jobs:
    - A `prepare` job which prepares the artifacts to deploy on the dev repo and uploads it as a workflow artifact.
    - A `deploy` job which checks out the `dist` version of the repo, cleans out most existing files, downloads and unpacks the prepared artifact and then commits the changes before pushing the branch/tag to the `dist` repo.
        The artifact upload/download mechanism with two jobs is used for two reasons:
        1. The way the environment in GH Actions is set up, is not all too friendly to checking out two repos at the same time and then being able to commit to the correct repo. (in other words: it's more secure)
            By splitting the deployment into two steps and uploading the artifact in between, we can be sure the commit ends up in the correct repo.
        2. To allow for examining the files in the artifacts for debugging.
5. Both jobs are explicitly prevented from being run on forks outside the Yoast organisation.
6. Both jobs contain a number of steps containing debug information, just in case.
7. In various places in both jobs, data from the original commit is used, such as the branch name, the committer and the commit message.
    This type of data should always be regarded as **untrusted** input and when these `github....` variables are used directly within the `run` context, they can lead to script injection and unintended execution of commands.
    To mitigate the risk of these type of script injection attacks, untrusted data is first set as a step-specific interim environment variable and only after that the environment variable (not the github variables directly) is used in the `run` context.
    This complies with the current best practices regarding defending against these type of attacks as per January 2022.
    A warning to this effect is at the top of the workflow to try and prevent people trying to be clever and "simplifying" the script.
    For more information, see: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions
8. In the `prepare` job, a GitHub access token is explicitly passed to the `setup-php` action to make it available to the workflow.
    This GH token has repo access to the `Yoast` (dev) organisation and is needed for the Composer install of YoastSEO (Free) via VCS.
    In contrast to the other workflows, this workflow _does_ use the Yoast dev repo to check out YoastSEO as it is needed for the creation of the artifact.
    This _does_ make this workflow susceptible to the Git bug - `BUG: refs/files-backend.c:2992: initial ref transaction called with existing refs` (which is avoided by the other workflow by using the `dist` repo).
    This is a risk which is unavoidable, but should also not cause too many problems in normal use for the below reasons. If the problem does occur, re-running the workflow (via the button in the web interface on GitHub) should hopefully resolve the issue.
    1. The `composer install` with scripts isn't run in the YoastSEO subdirectory.
    2. This workflow is only run on select branches (so less frequent).
    3. This workflow doesn't use a matrix of builds.
9. The `prepare` job is explicitly set up to use PHP 5.6 for the Composer installs and Node 14 for yarn and grunt.
    - PHP 5.6 is used as that's the minimum supported PHP version for this repo and there is nothing preventing a successfull install on PHP 5.6.
    - Node 14 is used for consistency with other scripts which really _need_ Node 14.
10. The `prepare` job will do an explicit `composer install` ahead of creating the artifact.
    The `grunt` jobs actually run `composer install` multiple times, both with `--no-dev` and with `--dev`. By running the `composer install` ahead of the `grunt` job, we can use the `ramsey/composer-install` action to cache the downloads Composer needs, thus warming up the cache and speeding up the build.
    Aside from that, for this particular workflow, it is extra important to have this step. This was also already in place in Travis with a note that builds would otherwise fail on a missing `yoastseo:link:vendor/yoast/wordpress-seo/packages/yoastseo` dependency.
11. In the `prepare` job, the update of the version number in the source code will only be executed for tags, not for manually triggered deploys or branch pushes as that would lead to weird/undesired versioning.
12. In the `deploy` job, the first two steps use some bash logic to create a short version of the commit sha and determine the branch which needs to be checked out and the commit title.
    While this type of logic, if simple, _can_ oftentimes be used inline, it is often more descriptive to extract it to a separate step and this also prevents repetitive logic in multiple places. More complex logic will, in most cases, needs such an interim step setting a variable anyway.
13. In the `deploy` job, a GitHub access token for the `dist` repo is used, which has been saved to the `dev` repos "secrets".
    This access token has been created with full access to the organisation and can commit to the `dist` repo.
    The use of GH access tokens and secrets this way is an advantage of using GH Actions and removes the need for creating an SSH key based on the `deploy_keys/wpseo_woo_deploy.enc` file as was previously done in Travis.
14. In the `deploy` job, by default the `actions/checkout` does a shallow clone.
    This means that other branches will not be known, so would always be created as new.
    By setting `fetch-depth` to `0`, a complete clone will be made, including all tags and branches.
    This will make the deploy slightly slower, but as the `dist` repos are nowhere near as huge as the `dev` repos, this should still be relatively limited.
    If this at some point would become problematic, we can have a think about using a "sensible number" for `fetch-depth` (something like `1000` or so).
    Refs:
    * https://github.com/actions/checkout#usage
    * https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches
15. In the `deploy` job, the commands which were previously executed in the `scripts/deploy_to_dist.sh` script have been transformed into job steps.
16. The commit for the deploy will be made by the "user" who created the tag.
17. In contrast to Travis, the _commit message_, will no longer _be_ the commit message of the commit which was pushed.
    Instead, the commit message will contain a range of information, _including the commit message of the commit which was pushed_, to be able to identify the commit and branch which triggered the deploy correctly.

Differences with the Travis implementation:
* The `master` branch is not included in the branch filtering as that branch was not included in the `deploy` specific branch filtering conditions.
    Tags pushed to `master` will still be deployed as those will be picked up via `push - tags`.
* Previously the deploy would run on the whatever PHP version would act as the default for the default Ubuntu image (`xenial` as per the top of the Travis script) on Travis. This has been changed to the minimum supported PHP version for this plugin (PHP 5.6).
* In the Travis script the `deploy` workflow also had a `WP_VERSION` set, but that looks to be unnecessary as WP isn't downloaded nor used during the deploy.

Includes making sure that Composer shell commands run from NPM/grunt use the Composer `--no-interaction` flag.

These workflows have been extensively tested already and need no further testing.

Aside from the builds run for this PR, you can find the results of various specific tests also on [the "Actions" page](https://github.com/Yoast/wpseo-woocommerce/actions).
For each job, it has been verified that the build(s) will actually show as failed when code has been altered to induce a fail condition.

Please do verify that the output from the `deploy` in the `dist` repo is as expected and thrown away this branch in the `dist` repo when merging this PR.


jrfnl:
Rebased for merge conflict

jrfnl:
Note: I've just done a force push with a small extra change to the `deploy` workflow as discussed on Slack.

diedexx (approved):

diedexx:
Resolved merge conflicts and cancelled the deploy action to prevent a push to yoast-dist. Merging as soon as all other checks succeed. and will remove the feature branch from yoast-dist too if that hasn't been done already

PR #754, opened 2022-03-03
Set minimum WP version to 5.8 and tested up to to 5.9

IreneStr:
## Context

* On January 25, 2022, WordPress 5.9 was released. We've decided to go back to supporting only the latest 2 WP versions.

## Summary

* Sets minimum required WordPress version to 5.8.

*

* Have WordPress 5.7 or lower installed. Be unable to update.
* See all github actions succeed.

*

## Impact check

*

## UI changes

## Documentation

## Quality assurance

Fixes #

### DIFF
--- a/.github/workflows/integrationtest.yml
+++ b/.github/workflows/integrationtest.yml
@@ -27,7 +27,7 @@ jobs:
       matrix:
         include:
           - php_version: '5.6'
-            wp_version: '5.6'
+            wp_version: '5.8'
             multisite: true
 
           - php_version: '7.0'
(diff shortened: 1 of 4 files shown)


jrfnl (commented):
@IreneStr LGTM - we could consider having one less build (I seem to remember we added one when the range got extended to more versions to make sure all versions were covered by test runs), but as GH Actions is fast anyway, this is absolutely fine ;-)

Only thing I'm thinking is maybe change `7.0` to `latest` and `7.3` to `trunk` ? 
That way we'd have a builds on the "low" and the "high" end of the PHP spectrum against both supported WP versions and one in the middle against the "upcoming release, may still fail".

IreneStr:
Makes sense, @jrfnl! I pushed a commit to make that change

PR #761, opened 2022-04-01
Add global identifiers for product variations

pls78:
## Context

*WooCommerce doesn't allows to spoecify global identifiers (e.g., GTIN12, ISBN, etc.) only for products. We want to add this functionality to products variations.

## Summary

* Adds global identifiers to products variations

*

Screenshot 2022-04-01 at 15 01 40

*

## Impact check

*

## UI changes

## Documentation

## Quality assurance

Fixes #

### DIFF
--- a/classes/woocommerce-yoast-ids.php
+++ b/classes/woocommerce-yoast-ids.php
@@ -1,0 +1,138 @@
+<?php
+/**
+ * WooCommerce Yoast SEO plugin file.
+ *
+ * @package WPSEO/WooCommerce
+ */
+
+/**
+ * Class WPSEO_WooCommerce_Yoast_Ids
+ */
+class WPSEO_WooCommerce_Yoast_Ids {
+
+	/**
+	 * The array of allowed identifier types.
+	 *
+	 * @var string[]
+	 */
+	protected $global_identifier_types = [
+		'gtin8'  => 'GTIN8',
+		'gtin12' => 'GTIN12 / UPC',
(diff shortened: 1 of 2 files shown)


classes/woocommerce-yoast-ids.php
```diff
@@ -0,0 +1,138 @@
...
+
+	/**
+	 * Save the $_POST values from the variation's ids div.
+	 *
+	 * @param int $variation_id The variation ID.
+	 *
+	 * @return bool Whether or not we saved data.
+	 */
+	public function save_data( $variation_id ) {
+		$nonce = filter_input( INPUT_POST, '_wpnonce_yoast_seo_woo' );
```
enricobattocchi, line 91: This is based on `classes/woocommerce-yoast-tab.php` but we're going to fix it because of this deprecation:
```
92 \| ERROR \| Using the filter_input() function without passing a filter means "FILTER_DEFAULT" will be used,
   \|       \| which is inherently insecure. (Yoast.PHP.FilterFunctions.DataNotFiltered)
```
I'd use the solution here from the start.

enricobattocchi (commented):
I'd fix one issue that we're going to fix in the "sibling class" for PHP 8.1 compatibility.
Also, some doubts about the design (see comment to issue in Jira)

classes/woocommerce-yoast-ids.php
pls78: I updated both the sanitization and the design issue.

leonidasmi:
CR : 👍

PR #773, opened 2022-05-26
DUPP-389 - DUPP-389-add-variation-global-ids-to-schema

pls78:
## Context

* We want to add variations global identifiers to a product's schema

## Summary

* Adds a feature to include global identifiers of product variations to a product's schema

* If a global identifier for a variation has not been specified, the value of that identifier is taken from the product (if specified)

* Create some variations of a product with the previous plugin version (or with `master` if you're a dev)
* Specify at least the price for each variation. **Note:** At least one variation needs to have a different price than the others for the proper `AggregateOffer` schema piece to be generated
* Add a couple of Product Global Indentifiers in the `Yoast SEO` tab of the WooCommerce metabox.
* Upgrade to the current RC (or checkout this branch if you're a dev)
* Check the frontend of the product you created and find the schema having `Product` as `@type` (Without saving it after you upgraded)
* Confirm that each `"@type": "Offer",` object has the Product Global Identifiers you had populated, with the value that you populated with.
* Also, validate the schema in the validators.
* Now, specify at least one **variation global identifier** for at least one variation and save the product:
image
* Also confirm that there is no ISBN variation global identifier.

* Check again the frontend of the product you created and find the schema having `Product` as `@type`
* In the `offers` array verify that each variation for which you specified a **variation global identifier** for has the corresponding attribute and value
* Also, validate the schema in the validators.
* Now edit the product as follows:
  * Delete the value of a global identifier you've previously specified for a variation
  * For the same global identifier, specify a value for the product in the `Yoast SEO` WooCommerce tab
* Now visit the product frontend post again, inspect the product schema and verify the variation whose global identifier you've previously deleted now has the product's global identifier value

* [X] QA should use the same steps as above.

## Impact check

*

## UI changes

## Documentation

## Quality assurance

Fixes [DUPP-389](https://yoast.atlassian.net/browse/DUPP-389)


classes/woocommerce-schema.php
```diff
@@ -510,6 +510,21 @@ protected function add_individual_offers( $product ) {
...
+			$variation_global_ids = get_post_meta( $variation['variation_id'], 'wpseo_variation_global_identifiers_values', true );
+			$product_global_ids   = get_post_meta( $product_id, 'wpseo_global_identifier_values', true );
+
+			if ( is_array( $variation_global_ids ) ) {
+				foreach ( $variation_global_ids as $global_id_name => $global_id_value ) {
+					if ( ! empty( $global_id_value ) ) {
+						$offer[ $global_id_name ] = $global_id_value;
+					}
+					elseif ( ! empty( $product_global_ids[ $global_id_name ] ) ) {
+						$offer[ $global_id_name ] = $product_global_ids[ $global_id_name ];
```
leonidasmi, line 523: Probably an `isset()` is needed here.

leonidasmi:
I see that if a variation has some global ids but not all, for the global ids that it doesnt have, it falls back to the respective product global id.

But if a variation has NO global ids, it doesnt fall back to the product global ids.

I assume that's a bug and not a feature 😅 I'll ask around, but until then we can go on with acceptance, because the rest of the CR is OK.

leonidasmi:
I'm also getting some warnings in the validator for the `isbn` attributes of the offers 🤔 
image

leonidasmi:
As per @jonoalderson 's advice, I'm gonna remove ISBN from this feature.

leonidasmi:
>But if a variation has NO global ids, it doesnt fall back to the product global ids.

It turned out that this is happening only when the variation wasnt saved after the new feature got introduced. Which means that if that's not fixed:
* People upgrading to this Yoast Woo version, they wont see their variations fall back to the global ids of the product unless they specifically save those variations again after upgrading

I'm gonna push a commit that fixes this, which means that:
* People upgrading will automatically see their "@type": "Offer", objects get populated with gtin8/gtin13/etc attributes as long as the Product has those IDs populated, without them saving those products again

maartenleenders:
Works!

PR #791, opened 2022-07-21
Send SKU data to worker on various variation manipulations

thijsoo:
## Context

* Lingo requires data to be scraped, prepared, and sent to the worker.

## Summary

* Scrapes data about SKUs from the DOM and sends it to the worker when needed.

* I have combined the identifier and SKU code in 1 file, because they both rely on some Jquery magic. And to make things easy I have added the small bit of code to also filter in the SKU data to the file.
* I have split up the PHP part.

Setup:
* Have Free checked out in` PC-139-enable-sku-assessment-for-woo` and built
* Go to Free and  copy the code from
packages\yoastseo\src\scoring\assessments\seo\ProductSKUAssessment.js to
to Woo's vendor `vendor\yoast\wordpress-seo\packages\yoastseo\src\scoring\assessments\seo\ProductSKUAssessment.js` file
* Build the vendor Free plugin
* Also build the WooCommerce plugin

(Test instructions are located in from https://github.com/Yoast/wordpress-seo/pull/18721):

* [X] QA should use the same steps as above.

*

## Impact check

*

## UI changes

## Documentation

## Quality assurance

* [X] I have tested this code to the best of my abilities

Fixes #
PC-139


js/src/yoastseo-woo-identifiers.js
```diff
@@ -11,14 +11,27 @@ const identifierKeys = [
...
 // Store to keep track of whether identifiersStore can be trusted or should be reloaded.
 let variantIdentifierDataIsValid = true;
 
+/**
+ * Checks whether the product has a global identifier
+ *
+ * @returns {boolean} Whether the product has a global identifier.
+ */
+function hasGlobalSKU() {
+	return Object.values( skuStore.global_sku ).some( identifier => identifier !== "" );
```
maartenleenders, line 32: I could be wrong but it does not seems like `global_sku` will ever be an array. So the `Array.some()` here is not necessary and you can directly check whether `global_sku !== ""` right?

classes/woocommerce-seo.php
```diff
@@ -1325,19 +1361,47 @@ private function localize_woo_identifiers() {
...
+		];
+
+		if ( WPSEO_WooCommerce_Utils::get_product_type( $product ) === 'variable' ) {
+			add_filter( 'woocommerce_hide_invisible_variations', [ $this, 'hide_invisible_variations' ] );
+			$variations = $product->get_available_variations();
+			remove_filter( 'woocommerce_hide_invisible_variations', [ $this, 'hide_invisible_variations' ] );
+
+			if ( ! empty( $variations ) ) {
+				$variation_ids                       = wp_list_pluck( $variations, 'variation_id' );
+				$identifiers['available_variations'] = $this->get_applicable_variations( $variations, 'display_price', $available_variations );
```
maartenleenders, line 1400: Since `$variations` is populated by `$product->get_available_variations`, why is there a need to check whether the display price is set for the variations?

maartenleenders: Never mind, I understand now. Available variations returns all variations. I misremembered.

classes/woocommerce-seo.php
```diff
@@ -604,6 +604,41 @@ public function set_yoast_columns_hidden_by_default( $current_screen ) {
...
+	 *
+	 * @param int    $variation_id   The post ID for the specific variation.
+	 * @param string $identifier_key The meta key that defines the needed identifier.
+	 *
+	 * @return string|\stdClass
+	 */
+	private function get_variant_identifier( $variation_id, $identifier_key ) {
+		$variation_identifier = get_post_meta( $variation_id, $identifier_key, true );
+
+		return ! empty( $variation_identifier ) ? $variation_identifier : new \stdClass();
```
maartenleenders, line 639: Some unfortunate type juggling going on here 
image
 but I think it's fine in this case. It would also be an option to simply return the `$variation_identifier` and let the caller handle possible undefined cases?

js/src/yoastseo-woo-identifiers.js
```diff
@@ -58,6 +71,26 @@ function doAllVariantsHaveIdentifier() {
...
+	if ( ! hasVariants() ) {
+		return false;
+	}
+
+	// Gather all identifier objects for each variation. Make sure each has at least one non-empty identifier.
+	return skuStore.available_variations.every(
+		availableVariationId => {
+			const variation = skuStore.variations[ availableVariationId ];
+			// Return true if at least one identifier is set.
+			return Object.values( variation ).some( variationIdentifier => variationIdentifier !== "" );
```
maartenleenders, line 89: Here as well, since you know you are looking for 1 sku per available variation, you can use the `availableVariationId` to get the `skuStore.variations[ availableVariationId ]` and check whether it's an empty object/string, or a filled string.

maartenleenders, line 89:
```suggestion
		availableVariationId => {
			const variationSku = skuStore.variations[ availableVariationId ];
			// Return true if variationSku is a nonempty string
			return typeof variationSku === "string" && variationSku.length > 0;
```

Something like this might work.

classes/woocommerce-seo.php
thijsoo: Yea I made it like this because in the identifiers you need an object. and for the sku there is only one. So we could also make an object with 1 key sku and then everything is a object.

maartenleenders: Yes I am aware of that. Let me see what I can do, I'll try both options 😉

maartenleenders:
CR & ACC ✅

PR #812, opened 2022-08-17
Update woocommerce-seo.php

Dekadinious:
Assign class instances to objects to enable removing of actions from other parts of the total codebase as per best practice.

## Context
* Using remove_action is not possible if there is no reference to a specific class.

## Summary

* Assign class instances to objects to enable removing actions

### DIFF
--- a/classes/woocommerce-seo.php
+++ b/classes/woocommerce-seo.php
@@ -74,8 +74,8 @@ class Yoast_WooCommerce_SEO {
 			// Move Woo box above SEO box.
 			add_action( 'admin_footer', [ $this, 'footer_js' ] );
 
-			new WPSEO_WooCommerce_Yoast_Tab();
-			new WPSEO_WooCommerce_Yoast_Ids();
+			$wpseo_woocommerce_yoast_tab = new WPSEO_WooCommerce_Yoast_Tab();
+			$wpseo_woocommerce_yoast_ids = new WPSEO_WooCommerce_Yoast_Ids();
 		}
 		else {
 			// Initialize schema & OpenGraph.


enricobattocchi:
Hi @Dekadinious, thanks for your contribution!
You definitely have a point about the things you're suggesting, and we want to merge the change, but we fail to see how it can help with `remove_action` if those variables are still not accessible from the outside of that method.
Can you provide us the code snippeT you were using to test the change? Thanks in advance!

Dekadinious:
@enricobattocchi That is a total oversight on my part during a hectic period! :) I don't know what you think would be the best course of action. Making them global variables, perhaps? I see some plugins return an array of classes initialized also...

enricobattocchi:
Hey @Dekadinious, no problem, I can relate :-) 
I think that to find the best strategy we'd need to understand the use case: which action in particular you were aiming to remove?

Dekadinious:
Yes, of course :)

See this code below. I have also provided the code from another plugin that makes the class available in a global.

`
//Remove product sync information box
global $woocommerce_gpf_di;
$woocommerceGpfAdmin = $woocommerce_gpf_di['WoocommerceGpfAdmin'];
remove_action('woocommerce_product_after_variable_attributes', [$woocommerceGpfAdmin, 'variation_input_fields'], 90);

//Remove Yoast SEO options
remove_action('woocommerce_product_after_variable_attributes', ['WPSEO_WooCommerce_Yoast_Ids', 'add_variations_global_ids'], 10);
`
So to be clear, removing the Yoast action is not working, and I have tried several things here, not just this.

enricobattocchi:
Hi @Dekadinious,
I see that WooCommerce is relying on a container that makes the classes available. Unfortunately we don't have such container handling the two classes you're targeting.
We definitely want to provide a way to disable the tab and the fields, but we'll do it in a separate issue by introducing new filters for that purpose. This is why I'm closing the issue for the moment.
They will be probably part of an upcoming version, though we don't have a scheduled release at the moment so I can't provide an expected date for it.

Thanks for your valuable suggestion!

PR #823, opened 2022-08-30
Fix SKU assessment not updating when variants are deleted after opening a saved product

iolandasequino:
## Context

*

## Summary

* Fixes an unreleased bug where the SKU and product identifiers assessment results would not update when opening a saved product with variants and deleting the variants.

* The bug was caused by the fact that if no variation elements are found, we get the variant data from the JavaScript object injected by the server. Since the variation elements are not immediately available on page load, this ensures that we still can retrieve the variants on page load. However, when we remove the variants after opening the page, the variation elements are also not found (because they were removed), but the data from the JS objected injected by the server is not updated without refreshing the page. So in that situation, we falsely think that the product still has variants and the assessment doesn't update. This is fixed by adding an additional check: we only get the initial data from the JS object injected by the server if the `data-total` attribute of the parent element of the variations is larger than 0. This means that there are variations, but the variation elements are not yet loaded. On the other hand, if there are no variation elements and the `data-total` attribute is 0, it means there are no variations, and an empty array is returned.

* Activate Yoast SEO Free, WooCommerce, and Yoast SEO: WooCommerce

#### Test the fix on saved products
* Create a new product with variants ([see documentation on how to add variants](https://yoast.atlassian.net/wiki/spaces/LIN/pages/2564686155/How+to+add+variants+to+a+product)). Do not save the product.
* Confirm that the SKU and product identifier assessments return orange bullets with the following feedback: `Product identifier/SKU: Not all your product variants have an identifier/a SKU. Include this if you can, as it will help search engines to better understand your content.`
* Add some text and a title, and save the product.
* Confirm that the assessment result stays the same.
* Delete each variant from the product individually.
* Confirm that the assessment results change to `SKU/Product identifier: Your product is missing a SKU/an identifier. Include this if you can, as it will help search engines to better understand your content.`
* Save the product and confirm that the assessment result stays the same.
* Add back the variants and confirm that the assessments return orange bullets with the following feedback: `Product identifier/SKU: Not all your product variants have an identifier/a SKU. Include this if you can, as it will help search engines to better understand your content.`
* Save the product again and confirm the feedback doesn't change.
* Delete the variants by using the `Delete all variations` option from the bulk action menu.
* Confirm that the assessment results change to `SKU/Product identifier: Your product is missing a SKU/an identifier. Include this if you can, as it will help search engines to better understand your content.`

#### Test with SKU and product identifier filled in
* Repeat the steps from the previous section, but this time fill in the SKU and product identifier of the variants when you add them.

#### Check whether updating still works on unsaved products
* Create a new product with variants and don't save it
* Confirm that the SKU and product identifier assessments return orange bullets with the following feedback: `Product identifier/SKU: Not all your product variants have an identifier/a SKU. Include this if you can, as it will help search engines to better understand your content.`
* Delete variants (either individually or by using the bulk action menu)
* Confirm that the assessment results change to `SKU/Product identifier: Your product is missing a SKU/an identifier. Include this if you can, as it will help search engines to better understand your content.`
* Repeat the above steps but this time fill in the SKU and product identifier of the variants.

* In addition to the acceptance test steps, re-test the feature using the steps from this issue https://yoast.atlassian.net/browse/PC-369 

## Impact check

*

## UI changes

## Documentation

## Quality assurance

Fixes https://yoast.atlassian.net/browse/PC-509

### DIFF
--- a/js/src/yoastseo-woo-identifiers.js
+++ b/js/src/yoastseo-woo-identifiers.js
@@ -97,8 +97,15 @@ function getInitialProductVariant( id ) {
 function getProductVariants() {
 	const variationElements = [ ...document.querySelectorAll( ".woocommerce_variation" ) ];
 
-	if ( variationElements.length === 0 ) {
-		// WooCommerce variations are not loaded yet, so try to use the initial data.
+	const variationsParentClass = document.querySelector( ".woocommerce_variations" );
+	const dataTotalAttribute = variationsParentClass.getAttribute( "data-total" );
+
+	/*
+	 * If no variation elements were found but the data-total attribute of the woo-commerce variations is not 0,
+	 * it means that there are variations but they are not yet loaded on page load. If that's the case, get the variations
+	 * from the JavaScript object injected by the server.
+	 */
+	if ( variationElements.length === 0 && dataTotalAttribute > 0 ) {
 		return Object.keys( wpseoWooIdentifiers.variations ).map( getInitialProductVariant );
 	}
 
--- a/js/src/yoastseo-woo-identifiers.js
+++ b/js/src/yoastseo-woo-identifiers.js
@@ -231,7 +231,11 @@ function registerEventListeners() {
 
 	// Listen for changes in the WooCommerce variations (e.g. adding or removing variations).
 	const variationsObserver = new MutationObserver( YoastSEO.app.refresh );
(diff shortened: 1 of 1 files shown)


iolandasequino:
CR: 💡 👏 
Acceptance: everything works as expected 🎉

PR #825, opened 2022-08-31
Fix SKU and product identifier assessments not showing up on a new product without variants

agnieszkaszuba:
## Context

* This [PR](https://github.com/Yoast/wpseo-woocommerce/pull/823) resulted in a console error and the assessments not showing up on a new product without variants. This is because it introduced code that always tries to get the attribute of the element with the `woocommerce_variations` class, even when that element doesn't exist. This PR fixes it so that we only get the attribute if the element exists.

## Summary

* Fixes an unreleased bug where the SKU and product identifier assessments did not show up on a new product without variants.

* If the element with the `woocommerce_variations` class doesn't exist, it means the product doesn't have variants, so we don't need to try to get the initial variant data.

* Activate Yoast SEO Free, WooCommerce, and Yoast SEO: WooCommerce
* Open a new product 
* For all the steps, pay extra attention that no console errors appear
* Confirm that the SKU and product identifier assessments return an orange bullet with the following feedback `SKU/Product identifier: Your product is missing a SKU/an identifier. Include this if you can, as it will help search engines to better understand your content.`
* Change the product to an external/affiliate product, and confirm that the result stays the same.
* Change the product to a grouped product and confirm that the assessments do not show up
* Change the product to a variable product, and confirm that the assessments return an orange bullet with the following feedback `SKU/Product identifier: Your product is missing a SKU/an identifier. Include this if you can, as it will help search engines to better understand your content.`
* Execute the testing steps from this PR to make sure that the fix from that PR still works: https://github.com/Yoast/wpseo-woocommerce/pull/823 

*

## Impact check

*

## UI changes

## Documentation

## Quality assurance

Fixes https://yoast.atlassian.net/browse/PC-526

### DIFF
--- a/js/src/yoastseo-woo-identifiers.js
+++ b/js/src/yoastseo-woo-identifiers.js
@@ -102,8 +102,8 @@ function getProductVariants() {
 
 	/*
 	 * If no variation elements were found but the data-total attribute on the variations parent element is larger than 0,
-	 * it means that there are variations but they are not yet loaded on page load. If that's the case, get the variations
-	 * from the JavaScript object injected by the server.
+	 * it means that there are variations but they are not yet loaded on page load. 
+	 * If that's the case, get the variations from the JavaScript object injected by the server.
 	 */
 	if ( variationElements.length === 0 && variationsParentElement ) {
 		const numberOfVariations = variationsParentElement.getAttribute( "data-total" );
--- a/js/src/yoastseo-woo-identifiers.js
+++ b/js/src/yoastseo-woo-identifiers.js
@@ -98,17 +98,18 @@ function getInitialProductVariant( id ) {
  */
 function getProductVariants() {
 	const variationElements = [ ...document.querySelectorAll( ".woocommerce_variation" ) ];
-
-	const variationsParentClass = document.querySelector( ".woocommerce_variations" );
-	const dataTotalAttribute = variationsParentClass.getAttribute( "data-total" );
+	const variationsParentElement = document.querySelector( ".woocommerce_variations" );
 
 	/*
-	 * If no variation elements were found but the data-total attribute of the woo-commerce variations is not 0,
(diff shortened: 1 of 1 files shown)


js/src/yoastseo-woo-identifiers.js
```diff
@@ -98,17 +98,18 @@ function getInitialProductVariant( id ) {
...
-
-	const variationsParentClass = document.querySelector( ".woocommerce_variations" );
-	const dataTotalAttribute = variationsParentClass.getAttribute( "data-total" );
+	const variationsParentElement = document.querySelector( ".woocommerce_variations" );
 
 	/*
-	 * If no variation elements were found but the data-total attribute of the woo-commerce variations is not 0,
+	 * If no variation elements were found but the data-total attribute on the variations parent element is larger than 0,
 	 * it means that there are variations but they are not yet loaded on page load. If that's the case, get the variations
 	 * from the JavaScript object injected by the server.
```
hdvos, line 106:
```suggestion
	 * it means that there are variations but they are not yet loaded on page load. 
	 * If that's the case, get the variations from the JavaScript object injected by the server.
```

Breaking the lines like this leads to a more readable comment.

js/src/yoastseo-woo-identifiers.js
```diff
@@ -98,17 +98,18 @@ function getInitialProductVariant( id ) {
...
 	 */
-	if ( variationElements.length === 0 && dataTotalAttribute > 0 ) {
-		return Object.keys( wpseoWooIdentifiers.variations ).map( getInitialProductVariant );
+	if ( variationElements.length === 0 && variationsParentElement ) {
+		const numberOfVariations = variationsParentElement.getAttribute( "data-total" );
```
hdvos, line 109: nice! A more clear variable name!

hdvos (approved):
No issues! Good job.
One small suggestion regarding line breaks. But it is not a blocker. 👍

hdvos:
All clear! 🎊

PR #830, opened 2022-09-14
Improve Offers / priceSpecification output

jdevalk:
## Context

* Improve `Offers` / `priceSpecification` output to match Google's recent changes to their guidelines.

## Summary

* Improves the Schema output for for `Offers` and `priceSpecification` to match Google's recent changes to their guidelines.
* Adds a feature to select a custom taxonomy for products to describe their `pattern` in the Schema output.

#### Product Schema changes
* Add some brands and manufacturers in Yoast SEO > WooCommerce SEO (the instructions are the same as per "Pattern" below)
* Create and publish some products (at least a simple one and a variable one)
* See that in the schema for `Product` > `offers` the `price` and `priceCurrency` properties have been moved inside `priceSpecification`
* See that the Schema has the correct `brand` and `manufacturer` properties: the former is of type `Brand`, the latter is of type "Organization"
```
"brand": {
  "@type": "Brand",
  "name": "Nike"
},
"manufacturer": {
  "@type": "Organization",
  "name": "Superga"
},
```
* check their Schema on the three tools we use
  * the changes should fix the issues in the Google Search Console about these two errors: `Invalid object type for field "brand"` and `Missing field "priceType"`
  * but we have to make sure we are not introducing errors in the other two validation tools.

#### Pattern
* visit Products > Attributes and add a new attribute for patterns
  * populate it with some terms
* visit Yoast SEO > WooCommerce SEO
  * see that there is a new "Pattern" dropdown
  * select the attribute we're dealing with and save the options
* edit a product
  * under Attributes, add the attribute you just created
  * populate it with some of the terms you created above 
  * Publish the product
* inspect the Schema of the product in the front-end
  * in the `Product` piece, you should see a "pattern" property which should be an array of the values you added to the Product

## UI changes

## Documentation

## Quality assurance

Fixes #https://yoast.atlassian.net/browse/IM-1962

### DIFF
--- a/classes/woocommerce-schema.php
+++ b/classes/woocommerce-schema.php
@@ -212,19 +212,20 @@ class WPSEO_WooCommerce_Schema {
 
 			// Add an @id to the offer.
 			if ( $offer['@type'] === 'Offer' ) {
-				$price                           = WPSEO_WooCommerce_Utils::get_product_display_price( $product );
 				$data['offers'][ $key ]['@id']   = YoastSEO()->meta->for_current_page()->site_url . '#/schema/offer/' . $product->get_id() . '-' . $key;
-				$data['offers'][ $key ]['price'] = $price;
 
-				$data['offers'][ $key ]['priceCurrency'] = get_woocommerce_currency();
+				$data['offers'][ $key ]['priceSpecification']['@type'] = 'PriceSpecification';
 
-				$data['offers'][ $key ]['priceSpecification']['@type']                 = 'PriceSpecification';
-				$data['offers'][ $key ]['priceSpecification']['valueAddedTaxIncluded'] = ( wc_tax_enabled() && WPSEO_WooCommerce_Utils::prices_have_tax_included() );
+				$data['offers'][ $key ]['seller'] = [ '@id' => YoastSEO()->meta->for_current_page()->site_url . '#organization' ];
 
-				// Remove priceSpecification price property from Schema output by WooCommerce.
-				unset( $data['offers'][ $key ]['priceSpecification']['price'] );
-				// Remove priceSpecification priceCurrency property from Schema output by WooCommerce.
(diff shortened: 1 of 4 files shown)


classes/woocommerce-schema.php
```diff
@@ -212,19 +213,20 @@ protected function filter_offers( $data, $product ) {
...
-				$price                           = WPSEO_WooCommerce_Utils::get_product_display_price( $product );
-				$data['offers'][ $key ]['@id']   = YoastSEO()->meta->for_current_page()->site_url . '#/schema/offer/' . $product->get_id() . '-' . $key;
-				$data['offers'][ $key ]['price'] = $price;
+				$data['offers'][ $key ]['@id'] = YoastSEO()->meta->for_current_page()->site_url . '#/schema/offer/' . $product->get_id() . '-' . $key;
 
-				$data['offers'][ $key ]['priceCurrency'] = get_woocommerce_currency();
+				$data['offers'][ $key ]['priceSpecification']['@type'] = 'PriceSpecification';
 
-				$data['offers'][ $key ]['priceSpecification']['@type']                 = 'PriceSpecification';
-				$data['offers'][ $key ]['priceSpecification']['valueAddedTaxIncluded'] = ( wc_tax_enabled() && WPSEO_WooCommerce_Utils::prices_have_tax_included() );
```
enricobattocchi, line 222: @jdevalk I'm afraid we're losing this part, the only other mention of `valueAddedTaxIncluded` is in the method for variations so simple products don't get this

jdevalk: Doesn’t Woo itself add that? If not we should.

enricobattocchi: Yeah, I'm looking deeper and it looks like it's doing it, I suppose it didn't until recently?

PR #832, opened 2022-09-15
PC-626 Fix the `color` and `pattern` output

enricobattocchi:
## Context

* We want to avoid the GSC tool to display a "Duplicate field" warning when `pattern` or `color` are set to an array of multiple element.

## Summary

* Fixes a bug where the Product Schema output would trigger a "Duplicate field" warning for the `color` property for non-variable products.
 
* the changelog lines mentions only the `color` property because the `pattern` one is being introduced in this version.

* Create two attributes, one for color and the other for pattern
* add some terms to both
* in the WooCommerce SEO settings, set the attributes to color and pattern
* Create a new product (either Simple, Grouped or External)
* under Attributes, add the attributes you just created
* populate each one with just **one** terms you created above
* Publish the product
* Inspect the schema and see that the `color` and `pattern` properties display a single value, e.g.:
```
"color": "green",
"pattern": "v-shape"
```
* check the schema validators that they don't complain about those properties
* edit the product and add terms to the attributes so they don't have only one value each
* update the product
* Inspect the schema and see that the `color` and `pattern` properties are not displayed anymore.
* Create a new **Variable** product
* under Attributes, add the attributes you just created
* populate each one with just **one** terms you created above
* Publish the product
* Inspect the schema and see that the `color` and `pattern` properties display a single value, e.g.:
```
"color": "green",
"pattern": "v-shape"
```
* edit the product and add terms to the attributes so they don't have only one value each
* update the product
* Inspect the schema and see that the `color` and `pattern` properties are now displayed as arrays:
```
"color": [ "green", "red" ],
"pattern": [ "v-shape", "cross" ],
```
* check the schema validators that they don't complain about those properties

*

## Impact check

*

## UI changes

## Documentation

## Quality assurance

Fixes [PC-626]

[PC-626]: https://yoast.atlassian.net/browse/PC-626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/classes/woocommerce-schema.php
+++ b/classes/woocommerce-schema.php
@@ -449,9 +449,11 @@ class WPSEO_WooCommerce_Schema {
 			if ( is_array( $terms ) ) {
 				// Variable products can have more than one color.
 				$is_variable_product = false;
-				foreach ( $this->data['offers'] as $offer ) {
-					if ( $offer['@type'] === 'AggregateOffer' ) {
-						$is_variable_product = true;
+				if ( isset( $data['offers'] ) ) {
+					foreach ( $this->data['offers'] as $offer ) {
+						if ( $offer['@type'] === 'AggregateOffer' ) {
+							$is_variable_product = true;
+						}
 					}
 				}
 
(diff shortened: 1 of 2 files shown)


leonidasmi:
CR + Acceptance is 🆗

classes/woocommerce-schema.php
```diff
@@ -447,12 +447,28 @@ private function add_color( $product ) {
 			$terms = get_the_terms( $product->get_id(), $schema_color );
 
 			if ( is_array( $terms ) ) {
-				$colors = [];
-				foreach ( $terms as $term ) {
-					$colors[] = strtolower( $term->name );
+				// Variable products can have more than one color.
+				$is_variable_product = false;
+				foreach ( $this->data['offers'] as $offer ) {
```
leonidasmi: What about some defensive coding here, so check if `isset( $this->data['offers']`?

leonidasmi:
CR of the later commits is 👍

PR #873, opened 2023-01-18
Add additional safeguard to make sure an indexable is actually found.

thijsoo:
## Context

* We want to add defensive coding to make sure indexables exists before we use them.

## Summary

* Adds defensive coding around opengraph data creation.

*

*

*

## Impact check

* This impacts the `open_graph_image_source` field in for when an open graph image is added to a product. So checking OG images op product pages is impacted.

## UI changes

## Documentation

## Quality assurance

## Innovation

* [X] This PR falls under an innovation project. I have attached the `innovation` label and noted the work hours.

Fixes #

### DIFF
--- a/classes/woocommerce-opengraph.php
+++ b/classes/woocommerce-opengraph.php
@@ -128,6 +128,6 @@ class WPSEO_WooCommerce_OpenGraph {
 		$indexable_repository = YoastSEO()->classes->get( 'Yoast\WP\SEO\Repositories\Indexable_Repository' );
 		$indexable            = $indexable_repository->find_by_id_and_type( $product_id, 'post' );
 
-		return $indexable->open_graph_image_source === 'set-by-user';
+		return ( $indexable && $indexable->open_graph_image_source === 'set-by-user' );
 	}
 }


classes/woocommerce-opengraph.php
```diff
@@ -128,6 +128,6 @@ protected function is_opengraph_image_set_by_user( $product_id ) {
 		$indexable_repository = YoastSEO()->classes->get( 'Yoast\WP\SEO\Repositories\Indexable_Repository' );
 		$indexable            = $indexable_repository->find_by_id_and_type( $product_id, 'post' );
 
-		return $indexable->open_graph_image_source === 'set-by-user';
+		return ( $indexable && $indexable->open_graph_image_source === 'set-by-user' );
```
leonidasmi, line 131: I think that this returning a false instead of an Indexable will not produce serious enough damage (aka no Fatal) for us to fix this here.

So, I'd rather not increase the scope of the release for a not as serious improvement as this one, what do you think @thijsoo ?

leonidasmi (commented):
If we decide not to do the above change, then we can skip this PR altogether, I don't see other instances either.

classes/woocommerce-opengraph.php
thijsoo: My thinking was if $indexable is false, and we try to ->open_graph_image_source wold that not cause an fatal?

leonidasmi: No, just a warning/notice:
image

It would be a fatal if we tried to call to a member function:
image

thijsoo: Ah makes sense!

PR #880, opened 2023-02-14
Load Woo worker in Elementor

agnieszkaszuba:
## Context

*

## Summary

* Loads Woo worker in Elementor

*

*

*

## Impact check

*

## UI changes

## Documentation

## Quality assurance

## Innovation

* [ ] This PR falls under an innovation project. I have attached the `innovation` label and noted the work hours.

Fixes #

### DIFF
--- a/js/src/yoastseo-woo-plugin.js
+++ b/js/src/yoastseo-woo-plugin.js
@@ -183,12 +183,20 @@ class YoastWooCommercePlugin {
 }
 
 /**
- * Initializes the Yoast Woo Commerce plugin.
+ * Loads the worker script in Elementor.
  *
- * @returns {YoastWooCommercePlugin} A new instance of the plugin class.
+ * @returns {void}
  */
-function initializeWooPlugin() {
-	return new YoastWooCommercePlugin(); // eslint-disable-line no-new
+function loadWorkerScriptInElementor() {
+	if ( typeof YoastSEO === "undefined" || typeof YoastSEO.analysis === "undefined" || typeof YoastSEO.analysis.worker === "undefined" ) {
+		return;
+	}
+
+	const worker = YoastSEO.analysis.worker;
+	const productDescription = getExcerpt();
+
+	worker.loadScript( wpseoWooL10n.script_url )
+	.then( () => worker.sendMessage( "initialize", { l10n: wpseoWooL10n, productDescription }, PLUGIN_NAME ) )
 }
 
 /**
@@ -199,7 +207,7 @@ if ( typeof YoastSEO !== "undefined" && typeof YoastSEO.app !== "undefined" ) {
 } else {
 	jQuery( window ).on( "elementor:init", () => {
 		window.elementor.on( "panel:init", () => {
-			addAction( "yoast.elementor.loaded", "yoast/yoast-woocommerce-seo/initialize-woo-plugin-in-elementor", initializeWooPlugin );
+			addAction( "yoast.elementor.loaded", "yoast/yoast-woocommerce-seo/load-worker-script-in-elementor", loadWorkerScriptInElementor );
 		} );
 	} );
 	jQuery( window ).on(
--- a/js/src/yoastseo-woo-plugin.js
+++ b/js/src/yoastseo-woo-plugin.js
@@ -196,7 +196,7 @@ function loadWorkerScriptInElementor() {
 	const productDescription = getExcerpt();
 
 	worker.loadScript( wpseoWooL10n.script_url )
-	.then( () => worker.sendMessage( "initialize", { l10n: wpseoWooL10n, productDescription }, PLUGIN_NAME ) )
+		.then( () => worker.sendMessage( "initialize", { l10n: wpseoWooL10n, productDescription }, PLUGIN_NAME ) );
 }
 
 /**
--- a/js/src/yoastseo-woo-plugin.js
+++ b/js/src/yoastseo-woo-plugin.js
@@ -2,6 +2,7 @@
 
 import { getExcerpt, addExcerptEventHandlers, isTinyMCEAvailable } from "./yoastseo-woo-handle-excerpt-editors";
 import { addFilter } from "@wordpress/hooks";
+import { addAction } from "@wordpress/hooks";
 import { dispatch } from "@wordpress/data";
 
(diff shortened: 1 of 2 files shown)


js/src/yoastseo-woo-plugin.js
```diff
@@ -181,12 +182,34 @@ class YoastWooCommercePlugin {
...
+ *
+ * @returns {void}
+ */
+function loadWorkerScriptInElementor() {
+	if ( typeof YoastSEO === "undefined" || typeof YoastSEO.analysis === "undefined" || typeof YoastSEO.analysis.worker === "undefined" ) {
+		return;
+	}
+
+	const worker = YoastSEO.analysis.worker;
+	const productDescription = getExcerpt();
```
hdvos, line 196: Do we really need productDescription already when loading the script? If so, I think it would be good to write a comment to document the reason.

js/src/yoastseo-woo-plugin.js
```diff
@@ -181,12 +182,34 @@ class YoastWooCommercePlugin {
...
+		.then( () => worker.sendMessage( "initialize", { l10n: wpseoWooL10n, productDescription }, PLUGIN_NAME ) );
+}
+
 /**
  * Adds eventlistener to load the Yoast WooCommerce plugin.
  */
 if ( typeof YoastSEO !== "undefined" && typeof YoastSEO.app !== "undefined" ) {
 	new YoastWooCommercePlugin(); // eslint-disable-line no-new
 } else {
+	jQuery( window ).on( "elementor:init", () => {
```
hdvos, line 208: This probably works. But maybe have a look if it can be done with JS-native methods (e.g. `window.addeventlistener`) as not to increase our dependency on other modules.

js/src/yoastseo-woo-plugin.js
```diff
@@ -181,12 +182,34 @@ class YoastWooCommercePlugin {
...
+function loadWorkerScriptInElementor() {
+	if ( typeof YoastSEO === "undefined" || typeof YoastSEO.analysis === "undefined" || typeof YoastSEO.analysis.worker === "undefined" ) {
+		return;
+	}
+
+	const worker = YoastSEO.analysis.worker;
+	const productDescription = getExcerpt();
+
+	worker.loadScript( wpseoWooL10n.script_url )
+		.then( () => worker.sendMessage( "initialize", { l10n: wpseoWooL10n, productDescription }, PLUGIN_NAME ) );
```
hdvos, line 199: HC suggested that here `worker.initialize()` might work as well instead of `worker.sendmessage`

hdvos (changes requested):
I have left some comments to have a look at.

Besides that, a few comments on the PR itself:
* In the context you refer to other issues that will fix the product description, product identifier, and SKU assessments. I think it would be good to already create these issues (if they are not there yet) and explicitely refer to them in this PR as to not overlook it.
* In context you mention that  _the product description, product identifier, and SKU assessments are still not working correctly._ I think it would be good to expand a bit on this. For example. Are they working, but give wrong results? Are they not visible?
* The impact check suggests that there are some user-facing changes in this PR. So maybe it is an enhancement instead of a non-user-facing change.

Besides from these details I think you did well on a complex issues.

js/src/yoastseo-woo-plugin.js
agnieszkaszuba: The `initialize` method of the worker requires the short product description, so with the current architecture we do need it. I'm not sure why it had been designed this way though. But if we want to get the product description in a different way in Elementor and in WP, then I think this approach would make sense. I'll make a note in the product description issue though to consider this.

hdvos:
I hope you are not too far in, but I am diving in the Elementor documentation for my current task. And I see that Elementor in it’s own documentation uses a lot of JQuery.

See for example: https://developers.elementor.com/docs/hooks/js/

So you might want to ignore my comment on jquery.

js/src/yoastseo-woo-plugin.js
agnieszkaszuba: I tried to implement this but couldn't get it to work. Maybe I misunderstand what exactly should be done. But I think since there are more uses of `worker.sendMessage( "initialize")` within the code, it might be best to create a technical debt issue where we potentially replace all of them.

This is what I tried to do:
- replaced this line with `.then( () => worker.initialize( { l10n: wpseoWooL10n, productDescription }, PLUGIN_NAME ) );`
- in `yoast-woo-worker.js`, added the `PLUGIN_NAME` parameter to the `initialize` method and removed this line from the `register` method: `this._worker.registerMessageHandler( "initialize", this.initialize.bind( this ), PLUGIN_NAME );`

After I add this change, the short product description does not appear anymore, so it looks like the method does not get executed.

hansjovis: I was wrong about what sendMessage does…

It does not serve as a replacement for `worker.initialize`, but only [serves as a way of sending custom messages](https://github.com/Yoast/wordpress-seo/blob/trunk/packages/yoastseo/src/worker/AnalysisWorkerWrapper.js#L244).

I assumed too much, without looking at the implementation :sweat_smile: .

agnieszkaszuba: Update: turns out that it's not possible to replace `worker.sendMessage( "initialize")` with `worker.initialize`. `worker.sendMessage` is used to register custom messages to the worker and cannot be used as a replacement for `worker.initialize`.

marinakoleva:
Final CR:

- Agreed that using a function here instead of a class is a much cleaner and nifty approach.  
- Technical choices section: tops marks for the detailed notes, appreciated!
- Updating `var` in the `yoastseo-woo-plugin.js` file to `const` or `let` will be left for a future refactoring project.

marinakoleva:
Acceptance testing:

- Testing instructions were expanded. This should ease RC testing too. 
- Two issues were created:
  - One concerning inconsistent behaviour of some assessments (happening both across and within the same content types and editors): [Fix inconsistent behaviour of Passive voice and Text length assessments](https://github.com/Yoast/wordpress-seo/issues/19937)
  - One concerning the slug: [Slug error in Product pages when WordProof Timestamp plugin is activated](https://github.com/Yoast/wordpress-seo/issues/19959)
- sd

PR #888, opened 2023-02-24
Implement bundle analyzer

FAMarfuaty:
## Context

*

## Summary

* Adds bundle analyzer plugin to Webpack config and makes `lodash` external dependency.

*

*

*

## Impact check

*

## UI changes

## Documentation

## Quality assurance

## Innovation

* [ ] This PR falls under an innovation project. I have attached the `innovation` label and noted the work hours.

Fixes #

### DIFF
--- a/config/webpack/webpack.config.default.js
+++ b/config/webpack/webpack.config.default.js
@@ -5,11 +5,14 @@ const { camelCaseDash } = require( "@wordpress/dependency-extraction-webpack-plu
 const UnminifiedWebpackPlugin = require( "unminified-webpack-plugin" );
 const CaseSensitivePathsPlugin = require( "case-sensitive-paths-webpack-plugin" );
 const { flattenVersionForFile } = require( "../grunt/lib/version.js" );
+const BundleAnalyzerPlugin = require( "webpack-bundle-analyzer" ).BundleAnalyzerPlugin;
+
 const webpack = require( "webpack" );
 
 const externals = {
 	yoastseo: "yoast.analysis",
 };
+let analyzerPort = 8888;
 
 /**
  * WordPress dependencies.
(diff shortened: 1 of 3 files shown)


config/webpack/webpack.config.default.js
```diff
@@ -58,6 +63,9 @@ const defaultConfig = {
 		} ),
 		new UnminifiedWebpackPlugin(),
 		new CaseSensitivePathsPlugin(),
+		new BundleAnalyzerPlugin( {
+			analyzerPort: analyzerPort++,
```
hansjovis, line 67: Can you document why we set/increase the analyzer port this way? Specifically as a comment in the code.

I noticed that we do it in WordPress SEO as well, but can't find a documented reason (e.g. in the git commit history).

My guess is that it allows us to run multiple analyzers next to each other, to be able to compare the bundle sizes of different branches / code bases without conflicting ports. But that would only be an educated guess.

If we don't know the reason to include it, I would remove it and just use:
```
new BundleAnalyzerPlugin()
```
since `8888` is already the default port.

config/webpack/webpack.config.default.js
```diff
@@ -5,11 +5,16 @@ const { camelCaseDash } = require( "@wordpress/dependency-extraction-webpack-plu
...
 const CaseSensitivePathsPlugin = require( "case-sensitive-paths-webpack-plugin" );
 const { flattenVersionForFile } = require( "../grunt/lib/version.js" );
+const BundleAnalyzerPlugin = require( "webpack-bundle-analyzer" ).BundleAnalyzerPlugin;
+
 const webpack = require( "webpack" );
 
 const externals = {
 	yoastseo: "yoast.analysis",
+	lodash: "window.lodash",
+	"lodash-es": "window.lodash",
```
hansjovis, line 15: These additions look good to me! 👍

hansjovis (changes requested):

config/webpack/webpack.config.default.js
FAMarfuaty: My only reason for doing it was because we also did it in WordPress SEO 😄 So I'll adjust it as you suggested 🙇🏽

FAMarfuaty:
Closed in favor of this PR: https://github.com/Yoast/wpseo-woocommerce/pull/889

PR #933, opened 2023-09-14
Use Translationspress instead of pulling translations at build time from translate.yoast.com

enricobattocchi:
## Context

* Since we are phasing out translate.yoast.com, we want to introduce the same mechanism which is in Premium and let users download the translations from Translationspress from inside their WP backend instead of bundling the .mo files

## Summary

* This PR introduces a new way of retrieving translations for Yoast SEO for WooCommerce, by utilizing the TranslationPress service. Instead of having to ship all translations with every release, we can now load the translations on a per-install basis, tailored to the user's setup. This means smaller plugin releases and less bloat on the user's server.

* There are no JS translations. This means that we don't need to bundle JSON files, nor we need a Grunt task for translations anymore
* There is no upgrade routine framework, nor the plugin version is stored in the DB, so I couldn't easily add an upgrade routine to pull the translations automatically. Users will need to visit the WP update page.

**Note:** TranslatePress only holds some languages at the moment. However, they might not be fully translated. 

#### Prerequisites for all tests
* Navigate to Settings -> General and install at least one additional language if you don't have any already.
* Checkout this branch and build accordingly (`composer install`, `yarn`, `grunt build:dev`)

#### Testing Grunt script
* Run `grunt build`.
* Check that the i18n related tasks don't run and no language files are bundled

#### Testing manually downloading language files
* Delete the language files in `/wp-content/languages/plugins/` (or delete the entire `plugins` directory itself)
* Navigate to Dashboard -> Updates and scroll down to the the translations update button at the bottom of the page.
* Click the update button. See that the translation files for all installed plugins are being downloaded.
* Check the final output of this downloading process 
   * You should see a number of lines which don't specify the plugin but only the language. They should be the WooSEO 
ones, there is a problem since the language domain `yoast-woo-seo` does not match the folder `wpseo-woocommerce`. 
   * there should one such line for each installed language (except of course en_US)
* Navigate to Yoast SEO -> Woocommerce SEO and see that the translations are working. Also test with additionally installed languages.

#### Testing outdated language files
* Navigate to `/wp-content/languages/plugins/` on your system.
* Edit yoast-woo-seo-<preferred language code>.po` and change the `PO-Revision-Date` to be earlier than the current value (i.e. change it by one day or one month or one year).
* Save your changes.
* Navigate to Dashboard -> Updates and scroll down to the the translations update button at the bottom of the page.
* See that new translations are available. 
* Click the update button.
   * You should see a number of lines which don't specify the plugin but only the language. They should be the Woo SEO 
ones, there is a problem since the language domain `yoast-woo-seo` does not match the folder `wpseo-woocommerce`. 
   * there should just one such line for the language you edited.
* Navigate to Yoast SEO -> Woocommerce SEO and see that the translations are working. Also test with additionally installed languages.

## Impact check

*

## UI changes

## Documentation

## Quality assurance

* [x] I have added unit tests to verify the code works as intended.
* [x] I have written this PR in accordance with my team's definition of done.

## Innovation

Fixes #


classes/integrations/woocommerce-translationspress.php
```diff
@@ -0,0 +1,209 @@
...
+		$installed_translations = \wp_get_installed_translations( 'plugins' );
+		$available_languages    = \get_available_languages();
+		foreach ( $translations[ $this->slug ]['translations'] as $translation ) {
+			if ( ! \in_array( $translation['language'], $available_languages, true ) ) {
+				continue;
+			}
+
+			if ( isset( $installed_translations[ $this->slug ][ $translation['language'] ] ) && $translation['updated'] ) {
+				$local  = new DateTime( $installed_translations[ $this->slug ][ $translation['language'] ]['PO-Revision-Date'] );
+				$remote = new DateTime( $translation['updated'] );
```
igorschoester, line 134: The DateTime constructor can throw an exception if the format is not correct.

Could be moved to private helper function for simplicity?

enricobattocchi:
> Besides the code comments: I think you can also remove the whole `languages` folder now. I.e. `languages/index.php`.

That's right! Done in a new commit

enricobattocchi:
About all the other comments: I discussed with @igorschoester and we agreed that it's OK to keep the class as it is at the moment since
- it comes from Premium where it's been working without issues since spring '21, so we can consider it well tested and safe
- the same class has been merged as it is in Local, so we can keep them consistent across all the plugins
- my main concern at the moment is to move to Translationspress so we can phase out translate.yoast.com

All the remarks from Igor are valid and interesting, but considering that there are still some other improvements regarding the translations in the pipeline, I'll collect them in a technical debt issue for the upcoming future.

thijsoo (approved):
CR + ACC 👍

PR #977, opened 2023-12-17
WPSEO_WooCommerce_Twitter::fallback_to_product_gallery_image(): fix broken condition

jrfnl:
## Context

* Twitter fallback image would be used, even when not applicable

## Summary

* Fixes a bug where Twitter fallback image would be used, even when not applicable.

`! $object->object_type === 'post'` is not the same as `$object->object_type !== 'post'`.

In the first case, the value of `$object->object_type` is cast to boolean before being compared to the string `post`, which means the condition would never match. Same goes for the second part of the condition.

I believe the code change I've made now fixes the condition to the _intended_ behaviour.

Includes tests.

* Go to `Yoast SEO` -> `Settings` -> `General` -> `Site features`
* In the `Social sharing` section, disable `Open Graph data`
* Without this PR:
  * edit `classes/woocommerce-twitter.php` and add the following line in line 42, between the `if` and the `return` statements:
  ```
  die( 'This method should only provides a fallback for products.' );
  ```
  * visit a post which doesn't have a Twitter image
  * verify you still see the post
* With this PR:
  * edit the same file as above, visit the same post and verify you now get  the message `This method should only provides a fallback for products.` instead of the post
* Perform the steps above also for a `page` 

*

## Impact check

*

## UI changes

## Documentation

## Quality assurance

* [ ] I have added unit tests to verify the code works as intended.
* [ ] I have written this PR in accordance with my team's definition of done.

## Innovation

Fixes #


pls78 (approved):
CR && Acc: ✅

PR #985, opened 2023-12-29
Upload combined coverage report for JS and PHP tests to Coveralls

diedexx:
## Context

* We want to make the impact a PR makes on our test coverage clear to everyone. We want to see if a change has an unexpected increase or decrease in test coverage and we want to have extra motivation/an extra reminder to increase coverage.

## Summary

* Runs JS tests in CI.
* Uploads test coverage of all our tested JavaScript packages to [Coveralls.io](https://coveralls.io/github/Yoast/wpseo-woocommerce)
* Separates Jest test configuration from the package.json.

* I specify collectCoverageFrom in our Jest config. Without this setting, Jest will only calculate coverage over tested files. We don't see if there is a file that has no tests covering them.
* I add all flags for all tests we do in the [carryforward](https://coveralls.io/better-monorepo-support) list. We only run JS tests if there are changes impacting JS, and the same applies to PHP. Because of that, we might not have coverage for the entire repo, skewing the coverage reporting. So, wo carry previous run results forward if those tests haven't been executed.
* I significantly lowered the global coverageThreshold. It did not reflect our current coverage at all, which caused test runs with `--coverage` to always fail. I took the current coverage percentage and rounded down to the decimal point.
* I remove redundant Jest config. The Jest defaults are the same or better.

* Check the build status of this PR. The JS test should be ran.
* Check [coveralls.io/github/Yoast/wordpress-seo-premium](https://coveralls.io/github/Yoast/wordpress-seo-premium): the last build for this PR should include coverage for JS as well as PHP.

* [n/a] QA should use the same steps as above.

*

## Impact check

* CI only

## UI changes

## Documentation

## Quality assurance

* [ ] I have added unit tests to verify the code works as intended.
* [ ] I have written this PR in accordance with my team's definition of done.

## Innovation


diedexx:
I've updated the Coveralls repo threshold to 44% (down from 45%) to account for the lower JS coverage for future PRs.

